PR #3, opened 2023-03-08
libddwaf v1.8.2

Julio-Guerra:
- [x] Update the libddwaf.a files to v1.8.2
- [x] Introduce `NewHandleFromRuleSet()` 1:1 with `NewHandle()` but now receiving a Go value instead, supposedly the parsed result of the WAF ruleset.
- [x] Remote `UpdateRulesData()` and corresponding tests.


waf_test.go
```diff
@@ -253,201 +250,9 @@ func TestActions(t *testing.T) {
...
-	test(map[string]interface{}{
-		"http.client_ip": "10.0.0.1",
-		"usr.id":         "moutix",
-	}, []string{"block_user", "block_ip"})
-}
-
 func TestAddresses(t *testing.T) {
 	defer requireZeroNBLiveCObjects(t)
-	expectedAddresses := []string{"my.first.input", "my.second.input", "my.third.input", "my.indexed.input"}
+	expectedAddresses := []string{"my.first.input", "my.second.input", "my.indexed.input", "my.third.input"}
```
Hellzy, line 255: Not sure I understand this one. Shouldn't `my.third.input` be third?

Hellzy (approved):
LGTM, thanks for taking care of the stripped .a for darwin 🙏
One small question

waf_test.go
Julio-Guerra: It's the new output of libddwaf v1.8
cc @Anilm3 if you want to have a look, but I am okay with it.

Julio-Guerra: Makes sens, we'll adjust the test next time it breaks, thanks.

PR #6, opened 2023-04-26
waf: add Update and UpdateFromRuleset bindings for ddwaf_update

eliottness:
Basic support for ddwaf_update


Julio-Guerra:
1. No longer compatible with `main`
2. To be part of the broader upgrade to the latest libddwaf API

PR #10, opened 2023-05-03
purego: wrappers

eliottness:
- [x] Add `libLoader` type
- [x] Add `wafLoader` type
- [x] Add type safe wrappers for all future calls to the waf
- [x] Add ELF Header test


lib_loader.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	return loader, symbols, nil
+}
+
+// getSymbol reloads a symbol from the shared library
+// if the uintptr symbol returned is 0 (nil), the error is a string given by dlerror()
+func (loader *libLoader) getSymbol(name string) (uintptr, error) {
+	return purego.Dlsym(loader.handle, name)
+}
```
Julio-Guerra, line 51: This method is not used.

```suggestion
```

lib_loader.go
```diff
@@ -0,0 +1,67 @@
...
+
+// syscall is the only way to make C calls with this interface.
```
Julio-Guerra, line 53: You mean the syscall ABI, right? It doesn't make an actual syscall in the end.

lib_loader.go
```diff
@@ -0,0 +1,67 @@
...
+// libLoader is created to wraps all interactions with the purego library
+type libLoader struct {
+	handle uintptr
+}
+
+// dlOpen open a handle for the shared library `name`.
+// the libLoader object is only a wrapped type for the linker handle
+// The `symbolsNeeded` is the list of symbols needed in the lib to load.
+// These symbols are returned via the map of symbols
+func dlOpen(name string, symbolsNeeded []string) (*libLoader, map[string]uintptr, error) {
```
Julio-Guerra, line 26: I suggest we make this API even more dynamic thanks to a new field tag `dlsym` so that we can directly fill in a Go struct with the expected symbols:

```go
type wafSymbols struct {
	Ddwaf_ruleset_info_free  uintptr `dlsym:"ddwaf_ruleset_info_free"`
	Ddwaf_init               uintptr `dlsym:"ddwaf_init"`
	Ddwaf_object_free        uintptr `dlsym:"ddwaf_object_free"`
	Ddwaf_destroy            uintptr `dlsym:"ddwaf_destroy"`
	Ddwaf_required_addresses uintptr `dlsym:"ddwaf_required_addresses"`
	Ddwaf_get_version        uintptr `dlsym:"ddwaf_get_version"`
	Ddwaf_context_init       uintptr `dlsym:"ddwaf_context_init"`
	Ddwaf_context_destroy    uintptr `dlsym:"ddwaf_context_destroy"`
	Ddwaf_result_free        uintptr `dlsym:"ddwaf_result_free"`
	Ddwaf_run                uintptr `dlsym:"ddwaf_run"`
}
```

and this function signature would become

```go
func dlOpen(name string, symbolsNeeded interface{}) (*libLoader, error)
// usage:
var symbols wafSymbols
lib, err := dlOpen("libddwaf.so", &symbols)
```

I think this will simplify the implementation of `newWafLoader` with a single source of symbol names `wafSymbols` in my example instead of three different Go values (the final struct, the map of symbol addresses, and the list of symbol names).

waf_loader.go
```diff
@@ -0,0 +1,151 @@
...
+	if err := os.WriteFile(file.Name(), libddwaf, 0400); err != nil {
+		return nil, fmt.Errorf("Error writing file: %w", err)
+	}
+
+	return file, nil
+}
+
+// newWafLoader loads the waf shared library and loads all the needed symbols
+func newWafLoader() (*wafLoader, error) {
+
```
Julio-Guerra, line 64:
```suggestion
```

Julio-Guerra (changes requested):

lib_loader.go
eliottness: Excellent idea, I'm on it :+1:

lib_loader.go
```diff
@@ -0,0 +1,81 @@
...
+
+// LibLoader is created to wraps all interactions with the purego library
+type LibLoader struct {
+	handle uintptr
+}
+
+// dlOpen open a handle for the shared library `name`.
+// the libLoader object is only a wrapped type for the linker handle
+// The `symbolsNeeded` is the list of symbols needed in the lib to load.
+// These symbols are returned via the map of symbols
```
Julio-Guerra, line 26: Deprecated comment :-)

lib_loader.go
```diff
@@ -0,0 +1,81 @@
...
+			}
+
+			loaderValue.Field(i).Set(reflect.ValueOf(symbol))
+			continue
+		}
+
+		if fieldType.Type == reflect.TypeOf(libLoader).Elem() {
+			loaderValue.Field(i).Set(reflect.ValueOf(libLoader).Elem())
+			foundHandle = true
+		}
```
Julio-Guerra, line 57: So this is why you made `LibLoader` public - I was wondering why.
I suggest we keep `LibLoader` private so that we can keep the go-libddwaf public API to its bare minimum without exposing the internals.
So you can change this expected type by `io.Closer`, by embedding `io.Closer` into the `wafLoader` struct.

lib_loader.go
```diff
@@ -0,0 +1,81 @@
...
+
+import (
+	"fmt"
+	"reflect"
+
+	"github.com/ebitengine/purego"
+)
+
+// LibLoader is created to wraps all interactions with the purego library
+type LibLoader struct {
```
Julio-Guerra, line 19: Sorry, I'm realising only now that this "loader" naming is not correct: this is not a loader, but rather its result.
I suggest you rename `{lib,waf}Loader` into `{lib,waf}Handle` or similar 🙏

lib_loader.go
```diff
@@ -0,0 +1,81 @@
...
+	handle uintptr
+}
+
+// dlOpen open a handle for the shared library `name`.
+// the libLoader object is only a wrapped type for the linker handle
+// The `symbolsNeeded` is the list of symbols needed in the lib to load.
+// These symbols are returned via the map of symbols
+func dlOpen(name string, loader any) error {
+
```
Julio-Guerra, line 28:
```suggestion
```

waf_loader.go
```diff
@@ -0,0 +1,134 @@
...
+		return nil, fmt.Errorf("test call in the C world failed: cannot load %s shared object. Reason: %v", file.Name(), err)
+	}
+
+	return &loader, nil
+}
+
+func (loader *wafLoader) tryCall() (err any) {
+	defer func() {
+		err = recover()
+	}()
```
Julio-Guerra, line 77: 💯

Julio-Guerra (changes requested):
Almost there 🙏

lib_loader.go
eliottness:
```suggestion
// dlOpen open a handle for the shared library `name`.
// the libLoader object is only a wrapped type for the linker handle
// the `loader` parameter must have tags in the form of `dlsym:<symbol_name>`
// dlOpen will fill the object with symbols loaded from the library
// the struct of type `loader` must have a field of type `LibLoader`
// to be able to close the handle later
```

lib_loader.go
eliottness: If I change the embedding into `io.Closer` in `wafLoader` I do not have access to the `syscall` function in wafLoader anymore. And I don't really know if it's ok to have an interface call in a very hot path like this

lib_loader.go
eliottness: This will be very confusing with the waf Handle returned by ddwaf_init

lib_loader.go
eliottness: done

lib_loader.go
eliottness: And I am not ok with the 'handle' name to begin with. What do you think of `libDl` and `wafDl` (dl as in dynamic library) ?

lib_dl.go
```diff
@@ -47,12 +49,12 @@ func dlOpen(name string, loader any) error {
...
 
-			loaderValue.Field(i).Set(reflect.ValueOf(symbol))
+			libValue.Field(i).Set(reflect.ValueOf(symbol))
 			continue
 		}
 
-		if fieldType.Type == reflect.TypeOf(libLoader).Elem() {
-			loaderValue.Field(i).Set(reflect.ValueOf(libLoader).Elem())
+		if fieldType.Type == reflect.TypeOf(dl) {
+			reflect.NewAt(reflect.TypeOf(dl), unsafe.Pointer(libValue.Field(i).UnsafeAddr())).Elem().Set(reflect.ValueOf(dl))
```
Julio-Guerra, line 57:
```suggestion
			// Bypass the fact the reflect package doesn't allow writing to private struct fields by directly writing to the field's memory address ourselves
			reflect.NewAt(reflect.TypeOf(dl), unsafe.Pointer(libValue.Field(i).UnsafeAddr())).Elem().Set(reflect.ValueOf(dl))
```

Julio-Guerra (approved):

PR #14, opened 2023-05-22
Purego wrappers fixes

eliottness:
* Update purego
* Use RTLD_DEFAULT to load libc symbols
* Add strlen and calloc symbols
* Disable zero-copy cstring
* Remove go-world callback
* Switch top-level wafObject allocation from go heap to calloc
* Remove foreign object global counter


Julio-Guerra (approved):
Approved but the removal of the dummy-and-desperate-C-allocation-tracker proved to be very efficient in the end and I will feel uncomfortable releasing dd-trace-go without it anymore. So I feel like we should move to the Go allocator asap now.

PR #15, opened 2023-05-31
Rework WAF Bindings 

eliottness:
Basically writing all the new implementation. Please look at the new `README.md` if you need an overview

```
Purego Copies (branche purego-v2):
BenchmarkOverall
BenchmarkOverall/16
BenchmarkOverall/16-10 	  263692	      4459 ns/op
BenchmarkOverall/64
BenchmarkOverall/64-10 	  255783	      4699 ns/op
BenchmarkOverall/128
BenchmarkOverall/128-10         	  238806	      4984 ns/op
BenchmarkOverall/256
BenchmarkOverall/256-10         	  213303	      5550 ns/op
BenchmarkOverall/512
BenchmarkOverall/512-10         	  180177	      6652 ns/op
BenchmarkOverall/1024
BenchmarkOverall/1024-10        	  128613	      8707 ns/op
BenchmarkOverall/2048
BenchmarkOverall/2048-10        	   97942	     12371 ns/op

CGO (branche main):
BenchmarkOverall
BenchmarkOverall/16
BenchmarkOverall/16-10 	  176250	      6634 ns/op
BenchmarkOverall/64
BenchmarkOverall/64-10 	  184581	      6558 ns/op
BenchmarkOverall/128
BenchmarkOverall/128-10         	  174375	      6719 ns/op
BenchmarkOverall/256
BenchmarkOverall/256-10         	  137481	      8316 ns/op
BenchmarkOverall/512
BenchmarkOverall/512-10         	  145016	      8254 ns/op
BenchmarkOverall/1024
BenchmarkOverall/1024-10        	  140026	      8677 ns/op
BenchmarkOverall/2048
BenchmarkOverall/2048-10        	  126534	      9512 ns/op
```


handle.go
```diff
@@ -0,0 +1,153 @@
...
+}
+
+// NewContext a new WAF context and increase the number of references to the WAF
+// handle. A nil value is returned when the WAF handle can no longer be used
+// or the WAF context couldn't be created.
+func (handle *Handle) NewContext() *Context {
+	// Handle has been released
+	if handle.contextCounter.Load() == 0 {
+		return nil
+	}
```
Julio-Guerra, line 86: The previous CAS implementation of the handle ref counter is still needed to avoid a race between the context creations and their WAF handle closing. `Load() == 0` isn't enough, and can become true right after it was false in a call to NewContext.

handle.go
```diff
@@ -0,0 +1,153 @@
...
+			Errors:  errorsMap,
+			Version: gostring(cRulesetInfo.version),
+		},
+	}, nil
+}
+
+// NewContext a new WAF context and increase the number of references to the WAF
+// handle. A nil value is returned when the WAF handle can no longer be used
+// or the WAF context couldn't be created.
+func (handle *Handle) NewContext() *Context {
```
Julio-Guerra, line 82: Sorry, it feels super weird in the end to see `NewContext` as a method 😅
I'd feel more comfortable reverting to a function to get back something more idiomatic.

Julio-Guerra (changes requested):
Quick first pass only mainly on the encoder and allocator.

context.go
```diff
@@ -0,0 +1,141 @@
...
+		dt := time.Since(now)
+		context.totalOverallRuntimeNs.Add(uint64(dt.Nanoseconds()))
+	}()
+
+	encoder := encoder{
+		stringMaxSize:    wafMaxStringLength,
+		containerMaxSize: wafMaxContainerSize,
+		objectMaxDepth:   wafMaxContainerDepth,
+	}
+	obj, err := encoder.Encode(addressesToData)
```
Julio-Guerra, line 64: After reviewing the purpose and usage of the allocator, I think this top-level `wafObject` is missing in the list of ref-counted Go values in `encoder.allocator`.

context.go
```diff
@@ -0,0 +1,141 @@
...
+// We store in there:
+// * A mutex to order calls to ddwaf_run()
+// * The handle it was created with
+// * The allocator to keep references to the go waf objects we sent
+// * A number of metrics
+type Context struct {
+	handle    *Handle
+	cContext  wafContext
+	mutex     sync.Mutex
+	allocator allocator
```
Julio-Guerra, line 29: why not `encoder` directly?

context.go
```diff
@@ -0,0 +1,141 @@
...
+		return nil, nil, err
+	}
+
+	// ddwaf_run cannot run concurrently and the next append write on the context state so we need a mutex
+	context.mutex.Lock()
+
+	defer context.mutex.Unlock()
+	defer context.allocator.append(encoder.allocator)
+
+	return context.run(obj, timeout, &encoder.allocator)
```
Julio-Guerra, line 75: I suggest you don't pass allocator here and simply call KeepAlive from here.

handle.go
Julio-Guerra: I added a test covering stressing the ref-counter in https://github.com/DataDog/go-libddwaf/pull/15/commits/8bb0d5b0310c6adebac3f7da214c324f9d5fb7a8 

It works like a charm and finds the issue :-)

handle.go
```diff
@@ -0,0 +1,153 @@
...
+type Handle struct {
+	cHandle        wafHandle
+	mutex          sync.RWMutex
+	contextCounter *atomic.Uint32
+	rulesetInfo    RulesetInfo
+}
+
+// NewHandle takes rules and the obfuscator config and return a new WAF to work with. The order of action is the following:
+// - Open the WAF library libddwaf.so using dlopen
+// - Encode the rule free form object as a ddwaf_object so send them to the WAF
```
Hellzy, line 30:
```suggestion
// - Encode the rule free form object as a ddwaf_object to send them to the WAF
```

handle.go
```diff
@@ -0,0 +1,153 @@
...
+func (handle *Handle) RulesetInfo() RulesetInfo {
+	return handle.rulesetInfo
+}
+
+// Addresses returns the list of addresses the WAF rule is expecting.
+func (handle *Handle) Addresses() []string {
+	return wafLib.wafRequiredAddresses(handle.cHandle)
+}
+
+// CloseContext call ddwaf_context_destroy and eventually ddwaf_destroy on the handle
```
Hellzy, line 107:
```suggestion
// CloseContext calls ddwaf_context_destroy and eventually ddwaf_destroy on the handle
```

handle.go
```diff
@@ -0,0 +1,153 @@
...
+func (handle *Handle) CloseContext(context *Context) {
+	wafLib.wafContextDestroy(context.cContext)
+	if handle.contextCounter.Dec() == 0 {
+		wafLib.wafDestroy(handle.cHandle)
+	}
+}
+
+// Close put the handle in termination state, when all the contexts are close the handle will be destroyed
```
Hellzy, line 115:
```suggestion
// Close puts the handle in termination state, when all the contexts are closed the handle will be destroyed
```

README.md
```diff
@@ -0,0 +1,120 @@
...
+- Call `ddwaf_run`
+- Decode the matches and actions
+
+### CGO-less C Bindings
+
+The main component used to build C bindings without using CGO is called [purego](https://github.com/ebitengine/purego). The flow of execution on our side is to embed the C shared library using `go:embed`. Then to dump it into a file, load it using `dlopen` and to load the symbols using `dlsym`. And finally to call them.
+
+⚠️ Keep in mind that __purego only works on linux/darwin for amd64/arm64 and so does go-libddwaf.__
+
+Another requirement because of this to have a FHS filesystem on your machine and, for linux, to provide `libc.so.6`, `libpthread.so.0` and `libm.so.6` as dynamic libraries.
```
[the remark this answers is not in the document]
eliottness:
```suggestion
Another requirement because of this to have a FHS filesystem on your machine and, for linux, to provide `libc.so.6`, `libpthread.so.0` and `libm.so.6`, `libdl.so.2` as dynamic libraries.
```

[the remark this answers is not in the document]
eliottness: 👍

README.md
```diff
@@ -0,0 +1,120 @@
...
+
+## Contributing usual pitfalls
+
+- Cannot dlopen twice in the app lifetime on OSX
+- `runtime.KeepAlive()` calls are here to prevent the GC from destroying objects too early
+- Since there is a stack switch between the go code and the C code. Usually the only C stacktrace you will ever get is from gdb
+- If a segfault happens during a call to the C code. The goroutine stacktrace which has done the call is the one annotated with `[syscall]`.
+- [GoLand](https://www.jetbrains.com/go/) does not support `CGO_ENABLED=0` (as of June 2023)
+- Keep in mind that we fully escape the type system. So if you send the wrong data it will segfaults in the best cases but not always!
```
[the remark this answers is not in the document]
eliottness:
```suggestion
- Keep in mind that we fully escape the type system. If you send the wrong data it will segfaults in the best cases but not always!
```

allocator.go
```diff
@@ -0,0 +1,107 @@
...
+		return
+	}
+
+	goArray := make([]byte, len(str))
+	copy(goArray, str)
+	allocator.stringRefs = append(allocator.stringRefs, goArray)
+
+	// This line firstly gets the value from the stringRef array. Then manage to get a *byte from the string.
+	// Then cast it into unsafe.Pointer then into uintptr
+	obj.value = uintptr(unsafe.Pointer(&goArray[0]))
```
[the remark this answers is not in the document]
eliottness: 😱 This makes my life really easier

context.go
```diff
@@ -0,0 +1,141 @@
...
+	timeoutCount atomic.Uint64
+}
+
+// Run encode the input data to send it to the WAF and returns the return values of ddwaf_run()
+// The parameters are the following:
+// * addressesToData: a map of key WAF addresses to the user input. The user input can be any go type.
+// * timeout: a timeout value to run the waf. If the threshold is exceeded, ddwaf_run will return without completing everything
+// The return values are the following:
+// * matches: A json string directly sent by the WAF containing all the data necessary for the backend
+// * actions: A array of string referencing which _blocking_ rules have been triggered
```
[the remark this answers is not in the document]
eliottness: Sure, let's change this
```suggestion
// * actions: A array of string referencing actions from triggered rules (`on_match` rule field)
```

encoder.go
```diff
@@ -0,0 +1,253 @@
...
+		length++
+	}
+
+	// Fix the size because we skipped map entries
+	obj.nbEntries = uint64(length)
+	return nil
+}
+
+// encodeMapKey takes a reflect.Value and a wafObject and returns a wafObject ready to be considered a map key
+// We use the function allocator.AllocWafMapKey to set store the key in the wafObject. But first we need
```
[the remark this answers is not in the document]
eliottness:
```suggestion
// We use the function allocator.AllocWafMapKey to store the key in the wafObject. But first we need
```

allocator.go
```diff
@@ -0,0 +1,104 @@
...
+func (allocator *allocator) AllocCString(str string) uintptr {
+	goArray := make([]byte, len(str)+1)
+	copy(goArray, str)
+	allocator.stringRefs = append(allocator.stringRefs, goArray)
+	goArray[len(str)] = 0 // Null termination byte for C strings
+
+	return uintptr(unsafe.Pointer(&goArray[0]))
+}
+
+func (allocator *allocator) AllocWafString(obj *wafObject, typ wafObjectType, str string) {
```
eliottness, line 51:
```suggestion
func (allocator *allocator) AllocWafString(obj *wafObject, str string) {
```

waf_disabled.go
```diff
@@ -13,15 +13,16 @@ import (
...
+type Handle struct{}
+
+type Context struct{}
 
 var errDisabledReason = errors.New(disabledReason)
 
+func InitWaf() error { return errDisabledReason }
+
+func CloseWaf() error { return errDisabledReason }
+
```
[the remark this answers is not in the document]
eliottness: Yes but we have to at least be able to compile with it since it's part of the API of the package

handle.go
```diff
@@ -0,0 +1,153 @@
...
+
+	emptyfree "github.com/DataDog/go-libddwaf/internal/empty-free"
+	"go.uber.org/atomic"
+)
+
+// Handle represents an instance of the WAF for a given ruleset.
+type Handle struct {
+	cHandle        wafHandle
+	mutex          sync.RWMutex
+	contextCounter *atomic.Uint32
```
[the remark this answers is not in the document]
eliottness: done

handle.go
eliottness: It's working :)

handle.go
```diff
@@ -0,0 +1,184 @@
...
+
+// NewHandle takes rules and the obfuscator config and returns a new WAF to work with. The order of action is the following:
+// - Open the WAF library libddwaf.so using dlopen
+// - Encode the rule free form object as a ddwaf_object to send them to the WAF
+// - Create a ddwaf_config object and fill the values
+// - Run ddwaf_init to get a handle from the waf
+// - Check for errors and streamline the ddwaf_ruleset_info returned
+func NewHandle(rules any, keyObfuscatorRegex string, valueObfuscatorRegex string) (*Handle, error) {
+
+	if err := InitWaf(); err != nil {
```
Julio-Guerra, line 52: Can we make `InitWaf` private?

handle.go
```diff
@@ -0,0 +1,184 @@
...
+	return handle.rulesetInfo
+}
+
+// Addresses returns the list of addresses the WAF rule is expecting.
+func (handle *Handle) Addresses() []string {
+	return wafLib.wafRequiredAddresses(handle.cHandle)
+}
+
+// CloseContext calls ddwaf_context_destroy and eventually ddwaf_destroy on the handle
+func (handle *Handle) CloseContext(context *Context) {
```
Julio-Guerra, line 123: Please make this method private

internal/empty-free/empty_free.go
```diff
@@ -3,9 +3,10 @@
...
-//go:build !cgo
+package empty_free
 
-package waf
+import "unsafe"
 
-var disabledReason = "cgo was disabled during the compilation and should be enabled in order to compile with the waf"
+//go:linkname _empty_free _empty_free
+var _empty_free byte
+var EmptyFreeFn uintptr = uintptr(unsafe.Pointer(&_empty_free))
```
Julio-Guerra, line 12: Let's rename it `NoopFreeFn`:
```suggestion
var NoopFreeFn uintptr = uintptr(unsafe.Pointer(&_empty_free))
```

handle.go
```diff
@@ -0,0 +1,184 @@
...
+		limits: wafConfigLimits{
+			maxContainerDepth: wafMaxContainerDepth,
+			maxContainerSize:  wafMaxContainerSize,
+			maxStringLength:   wafMaxStringLength,
+		},
+		obfuscator: wafConfigObfuscator{
+			keyRegex:   cgoRefs.AllocCString(keyObfuscatorRegex),
+			valueRegex: cgoRefs.AllocCString(valueObfuscatorRegex),
+		},
+		freeFn: emptyfree.EmptyFreeFn,
```
Julio-Guerra, line 168:
```suggestion
		// Prevent libddwaf from freeing our Go-memory-allocated ddwaf_objects 
		freeFn: emptyfree.EmptyFreeFn,
```

handle.go
eliottness: Yes

waf.go
```diff
@@ -3,975 +3,44 @@
...
-	}
-}
-
-// Increment the reference counter.
-// CAS spin-loop implementation in order to enforce +1 cannot happen when 0 has
-// been reached.
-func (c *atomicRefCounter) increment() uint32 {
-	return c.add(1)
+func Health() error {
+	return InitWaf()
```
Julio-Guerra, line 41: As of the current state of `InitWaf`, `Health()` can currently fail if things like `os.Remove` fails, while we must be more permissive and let using the libddwaf as long as it is usable. OTOH, I think all the errors are worth noting via the dd-trace-go logger, which we can't access from here.

For this reason, I suggest we move to a slightly different API where we don't loose the details:
1. Rename `InitWaf()` into `Load()`, and `CloseWaf()` into `Close()`.
2. Change the API of `Load()` into `func Load() (ok bool, err error)` so that we can let the caller can know if libddwaf is usable or not thanks to `ok`:
    1. When `ok` is true, a non-nil error is acceptable to me in order to return such `os.Remove()` errors for instance that we can log in the dd-trace-go side.
    2. When `ok` is false (ie. not usable), the error must state why, and so such `os.Remove()` errors shouldn't take precedence over the actual libddwaf-related errors.

I'm ok doing such changes in a dedicated separate PR.

waf_dl.go
```diff
@@ -47,18 +46,19 @@ func dumpWafLibrary() (*os.File, error) {
...
 	file, err := dumpWafLibrary()
 	if err != nil {
 		return nil, err
 	}
 
 	defer func() {
-		file.Close()
-		os.Remove(file.Name())
+		err = file.Close()
+		err = os.Remove(file.Name())
```
Julio-Guerra, line 57: I think we should ignore them, or at least keep the error if already not nil.

handle.go
Julio-Guerra: Closing this for now as I have new ideas around Health vs InitWaf basically coming in my pending review.

handle.go
eliottness: done

allocator.go
eliottness: done

waf.go
eliottness: Sure let there be a Jira ticket

context.go
```diff
@@ -0,0 +1,143 @@
...
+}
+
+// Run encodes the input data to send it to the WAF and returns the return values of ddwaf_run()
+// The parameters are the following:
+// * addressesToData: a map of key WAF addresses to the user input. The user input can be any go type.
+// * timeout: a timeout value to run the waf. If the threshold is exceeded, ddwaf_run will return without completing everything
+// The return values are the following:
+// * matches: A json string directly sent by the WAF containing all the data necessary for the backend
+// * actions: A array of string referencing actions from triggered rules (`on_match` rule field)
+// * err: a RunError error from the error codes the WAF returns
```
Julio-Guerra, line 52: Making it a bit higher level and better looking for godoc:

```suggestion
// Run encodes the given addressesToData values and runs them against the WAF rules within the given
// timeout value. It returns the matches as a JSON string (usually opaquely used) along with the corresponding
// actions in any. In case of an error, matches and actions can still be returned, for instance in the case of a
// timeout error. Errors can be tested against the RunError type.
```

ctypes.go
```diff
@@ -124,3 +132,45 @@ func cstring(name string) *byte {
...
+func sliceToUintptr[T any](arg []T) uintptr {
+	return (*reflect.SliceHeader)(unsafe.Pointer(&arg)).Data
+}
+
+func stringToUintptr(arg string) uintptr {
+	return (*reflect.StringHeader)(unsafe.Pointer(&arg)).Data
+}
+
+var alwaysFalse bool
+var escapeSink any
```
Julio-Guerra, line 163:
```suggestion
// keepAlive() globals
var (
	alwaysFalse bool
	escapeSink any
)
```

ctypes.go
```diff
@@ -124,3 +132,45 @@ func cstring(name string) *byte {
...
+
+// keepAlive is a copy of runtime.KeepAlive
+// keepAlive has 2 usages:
+// It forces the deallocation of the memory to take place later than expected (just like runtime.KeepAlive), and
+// it forces the memory to be on the heap by calling a variadic parameter function which calls to go:linkname function.
+// Of course the if is always true and the function always returns but the compiler does not know this
```
Julio-Guerra, line 169:
```suggestion
// - It forces the deallocation of the memory to take place later than expected (just like runtime.KeepAlive)
// - It forces the given argument x to be escaped on the heap by saving it into a global value (Go doesn't provide a standard way to do it as of today)
// It is implemented so that the compiler cannot optimize it.
```

waf.go
eliottness: done

context.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+	obj, err := encoder.Encode(addressesToData)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// ddwaf_run cannot run concurrently and the next append write on the context state so we need a mutex
+	context.mutex.Lock()
+	defer context.mutex.Unlock()
+	defer context.cgoRefs.append(encoder.cgoRefs)
```
Julio-Guerra, line 67:
```suggestion
	// Save the Go pointer references to addressesToData that were referenced by the encoder
	//  into C ddwaf_objects. libddwaf's API requires to keep this data for the lifetime of the ddwaf_context.
	defer context.cgoRefs.append(encoder.cgoRefs)
```

context.go
```diff
@@ -0,0 +1,133 @@
...
+
+	return matches, actions, err
+}
+
+// Close calls handle.closeContext which calls ddwaf_context_destroy and maybe also close the handle if it in termination state.
+func (context *Context) Close() {
+	// Needed to make sure the garbage collector does not throw the values send to the WAF earlier than necessary
+	keepAlive(context.cgoRefs.stringRefs)
+	keepAlive(context.cgoRefs.arrayRefs)
+	context.handle.closeContext(context)
```
Julio-Guerra, line 121: 1. I don't see any value for `closeContext` which spreads to context closure into an handle method
2. The `keepAlive` calls don't make sense to me here, because nothing special happens before.

So here is my suggestion:
```suggestion
	defer context.handle.closeContext(context)
	// Needed to make sure the garbage collector does not throw the values send to the WAF earlier than necessary
	wafLib.wafContextDestroy(context.cContext)
	// Keep the Go pointer references until the end of the context
	keepAlive(context.cgoRefs)
	// The context is no longer used so we can try releasing the Go pointer references asap by nulling them
	context.cgoRefs = cgoRefPool{}
```

If applied, renaming `closeContext` sounds better too.

ctypes.go
```diff
@@ -124,3 +132,48 @@ func cstring(name string) *byte {
 	copy(b, name)
 	return &b[0]
 }
+
+// cast is used to centralized unsafe use C allocated pointer.
```
Julio-Guerra, line 136:
```suggestion
// cast is used to centralize unsafe use C allocated pointer.
```

ctypes.go
```diff
@@ -124,3 +132,48 @@ func cstring(name string) *byte {
...
+// We take the address and then dereference it to trick go vet from creating a possible misuse of unsafe.Pointer
+func cast[T any](ptr uintptr) *T {
```
Julio-Guerra, line 138: If you want to give it a try, I think this is the solution to accept pointers and knowing the pointer type:

```go
func cast[T any, PT interface { *T }](ptr uintptr) PT
```

Julio-Guerra (commented):
New pass (still not finished 😬)

Hellzy (approved):
I think we've iterated enough for a first version, this looks okay to me. Great job 💪
Let's address the last few comments and wait for @Julio-Guerra's approval 👍

context.go
eliottness: I was against the idea because of owernship. Just like why NewContext() is within the handle scope, I think that destroying the context should also be done via the handle. This idea already became a little flaky after you told me to change NewContext() from a method to a function but yeah

ctypes.go
eliottness: image

context.go
Julio-Guerra: Part of this proposal still apply even with closeContext:
```suggestion
	defer context.handle.closeContext(context)
	// Keep the Go pointer references until the end of the context
	keepAlive(context.cgoRefs)
	// The context is no longer used so we can try releasing the Go pointer references asap by nulling them
	context.cgoRefs = cgoRefPool{}
```

handle.go
```diff
@@ -0,0 +1,183 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("could not encode the WAF ruleset into a WAF object: %w", err)
+	}
+
+	config := newConfig(&encoder.cgoRefs, keyObfuscatorRegex, valueObfuscatorRegex)
+	cRulesetInfo := new(wafRulesetInfo)
+
+	cHandle := wafLib.wafInit(obj, config, cRulesetInfo)
+	if cHandle == 0 {
+		return nil, errors.New("could not instanciate the WAF")
```
Julio-Guerra, line 71:
```suggestion
		return nil, errors.New("could not instantiate the WAF")
```

Julio-Guerra (changes requested):

handle.go
```diff
@@ -0,0 +1,183 @@
...
+}
+
+// Close puts the handle in termination state, when all the contexts are closed the handle will be destroyed
+func (handle *Handle) Close() {
+	// There are still Contexts that are not closed
+	if handle.addRefCounter(-1) > 0 {
+		return
+	}
+
+	wafLib.wafDestroy(handle.cHandle)
```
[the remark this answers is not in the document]
eliottness: Do you want me to move it ? I would need to rename everything since it does not only decrementReferences since it also closes 👀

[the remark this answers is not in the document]
Julio-Guerra: I'm good, simple comment to emphasis it.

Julio-Guerra (approved):

PR #17, opened 2023-07-03
Public API: Load()

Julio-Guerra:
Sorry, this is a big-bang PR implementing:
- `func Load() (ok bool, err error)` to ease the integration in dd-trace-go so that we can distinguish the different possible error cases:
   - Panic errors that must be used to gracefully abort the caller and prevent it from continuing. Such errors are represented by the new `PanicError` error type.
   - Unsupported target OS and/or architecture, which must be handled as an informative error, is worth logging to the user, and prevents the caller from continuing. This error is represented by the new `UnsupportedTarget` error type.
   - Every other unexpected error that can possible happen is returned as-is, but `ok` is `false`, so that the caller knows it cannot use the package.
   - Informative errors, about internal non-critical failures, can also happen (eg. if os.Remove fails removing the temporary files), and are returned as-is, but with `ok` being `true` so that the caller can log the error and continue its startup.
   - The dd-trace-go integration of this PR can be found here https://github.com/DataDog/dd-trace-go/pull/2090
   - Note that `Health()` has been kept for the sake of simplicity health-checking cases such as in the tests, where it returns the `Load()` error only if `ok` is false, meaning that go-libddwaf is not healthy.
 - Simplified build tags, removed everywhere it was possible, down to purego which isn't properly stubbed for unsupported targets and which we stub ourselves at our level in the new `waf_dl_unsupported.go` file.


waf_test.go
```diff
@@ -23,11 +23,13 @@ import (
 )
 
 func init() {
-	InitWaf()
+	Load()
```
eliottness, line 26:
```suggestion
	if err =: Load(); err != nil {
	    panic(err)
	}
```

eliottness (approved):
LGTM (except maybe for the little comment i left)

safe.go
```diff
@@ -0,0 +1,72 @@
...
+// PanicError is an error type wrapping a recovered panic value that happened
+// during a function call. Such error must be considered unrecoverable and be
+// used to try to gracefully abort. Keeping using this package after such an
+// error is unreliable and the caller must rather stop using the library.
+// Examples include safety checks errors.
+type PanicError struct {
+	// The function that was given to `Call()`.
+	In func() error
+	// The recovered panic value while executing `In()`.
+	Err error
```
Hellzy, line 25: Is there a need for those to be public? Seems like we could "leak" some internal functions through this which seems like an odd behaviour.
Also, what about having newPanicError embed the function's name instead, since it seems to be the purpose of embedding the function here? It's less "lazy" but we don't really need to be lazy for this since a panic occurred.

waf.go
```diff
@@ -3,41 +3,131 @@
...
+		return err
+	}
+	return nil
 }
 
+var wafVersion string
+
+// Version returns the version returned by libddwaf.
+// It relies on the dynamic loading of the library, which can fail and return
+// an empty string.
```
Hellzy, line 129:
```suggestion
// an empty string or the previously loaded version, if any.
```

waf_dl.go
```diff
@@ -44,42 +43,53 @@ func dumpWafLibrary() (*os.File, error) {
...
 	defer func() {
-		// TODO(eliott.bouhana): find a way to send to the tracer some non-fatal errors while opening the waf libg
-		file.Close()
-		os.Remove(file.Name())
+		rmErr := os.Remove(fName)
+		if err == nil {
+			err = rmErr
+		} else {
+			// TODO: rely on errors.Join() once go1.20 is our min supported Go version
+			err = fmt.Errorf("%w; along with an error while removing %s: %v", err, fName, rmErr)
```
Hellzy, line 63: What if `err` is set but there was no error returned by `os.Remove(fName)`? Seems to me like the message will be inaccurate and we could be printing a nil `rmErr`.

waf_dl.go
```diff
@@ -44,42 +43,53 @@ func dumpWafLibrary() (*os.File, error) {
...
+		} else {
+			// TODO: rely on errors.Join() once go1.20 is our min supported Go version
+			err = fmt.Errorf("%w; along with an error while closing the shared libddwaf library file: %v", err, closeErr)
+		}
+	}()
 
-	// try a call to the waf to make sure everything is fine
-	if err := waf.tryCall(); err != nil {
-		return nil, fmt.Errorf("test call in the C world failed: cannot load %s shared object. Reason: %v", file.Name(), err)
+	// Try a call to the waf to make sure everything is fine
```
Hellzy, line 81:
```suggestion
	// Try calling the waf to make sure everything is fine
```

Hellzy (commented):
Couple remarks, nothing too mean

safe.go
Julio-Guerra: Good points yes, it was done in 82c1a4d32b97f515876efc61490831b7d76120bb
I kept the `Err` field public as this is a wrapper error type where the usual "paradigm" is to give access to the underlying wrapped error.

waf_dl.go
Julio-Guerra: Good catch, that was a mistake indeed, fixed in 73948c9
A shame we can hardly test those lines, unless relying on interface abstractions to mock them in tests -_-

waf.go
```diff
@@ -3,41 +3,131 @@
...
 		}
-
 		wafVersion = wafLib.wafGetVersion()
 	})
 
-	return wafErr
+	return wafLib != nil, wafErr
 }
 
+// Health returns an error when this package is not usable state describing
```
Hellzy, line 115:
```suggestion
// Health returns an error when this package is not in a usable state describing
```

Hellzy (approved):
LGTM 👍

PR #18, opened 2023-07-05
Public API: Replace Health() by SupportsTarget()

Julio-Guerra:
While integrating go-libddwaf v1.3.0 into dd-trace-go, we figured out that `Load()` was doing too much for the case of remote activation of appsec when in a disabled state, where we don't want to load libddwaf at all until appsec actually starts up remote activation order.

For this reason, we decided to make `Health()` useful again by replacing it with `SupportsTarget()` in order to be able to only check if we are on a supported target or not. This allows us to properly check for proper support of libddwaf, apart from the libddwaf loading, to do it early in the appsec startup and be able to properly log the information to the user.

### DIFF
--- a/waf.go
+++ b/waf.go
@@ -113,7 +113,7 @@ func Load() (ok bool, err error) {
 }
 
 // SupportsTarget returns true and a nil error when the target host environment
-// is not supported, such as with a non-supported operating-system.
+// is supported by this package and can be further used.
 // Otherwise, it returns false along with an error detailing why.
 func SupportsTarget() (bool, error) {
 	return supportsTarget()
--- a/waf.go
+++ b/waf.go
@@ -112,14 +112,11 @@ func Load() (ok bool, err error) {
 	return wafLib != nil, wafErr
 }
 
-// Health returns an error when this package is not in a usable state describing
-// the reason why.
-func Health() error {
-	ok, err := Load()
-	if !ok {
-		return err
-	}
-	return nil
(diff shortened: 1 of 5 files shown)


waf_dl_unsupported.go
```diff
@@ -15,8 +15,10 @@ import (
 
 type wafDl struct{}
 
+var UnsupportedTargetErr = &UnsupportedTargetError{fmt.Errorf("the target operating-system %s or architecture %s are not supported", runtime.GOOS, runtime.GOARCH)}
```
eliottness, line 18: I don't think we need this variable to be public

eliottness (approved):

waf.go
```diff
@@ -112,14 +112,11 @@ func Load() (ok bool, err error) {
...
-// the reason why.
-func Health() error {
-	ok, err := Load()
-	if !ok {
-		return err
-	}
-	return nil
+// SupportsTarget returns true and a nil error when the target host environment
+// is not supported, such as with a non-supported operating-system.
+// Otherwise, it returns false along with an error detailing why.
```
Julio-Guerra, line 117:
```suggestion
// SupportsTarget returns true and a nil error when the target host environment
// is supported by this package and can be further used.
// Otherwise, it returns false along with an error detailing why.
```

waf_dl_unsupported.go
Julio-Guerra: Fixed by 28f73a9

Hellzy (approved):

PR #19, opened 2023-07-17
api: explicitly disable future go versions

Julio-Guerra:
purego is panic'ing on go1.21's release candidate, which will be released in August.
With this PR, we will start enforcing the set of Go versions we know we support and explicitly disable future Go versions until we tested them ourselves. As a result, Go 1.21 is disabled for now until a fix is figured out.

### DIFF
--- a/waf_unsupported_go.go
+++ b/waf_unsupported_go.go
@@ -4,6 +4,7 @@
 // Copyright 2016-present Datadog, Inc.
 
 // Supported OS/Arch but unsupported Go version
+//           Supported OS        Supported Arch     Bad Go Version
 //go:build (linux || darwin) && (amd64 || arm64) && go1.21
 
 package waf
--- a/waf_unsupported_go.go
+++ b/waf_unsupported_go.go
@@ -3,11 +3,14 @@
 // This product includes software developed at Datadog (https://www.datadoghq.com/).
 // Copyright 2016-present Datadog, Inc.
 
-//go:build darwin && arm64 && !go1.21
+// Supported OS/Arch but unsupported Go version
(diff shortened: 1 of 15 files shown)


waf_unsupported_target.go
```diff
@@ -0,0 +1,16 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-present Datadog, Inc.
+
+// Unsupported target OS or architecture on a supported Go version
+//go:build (!linux && !darwin) || (!amd64 && !arm64 && !go1.21)
```
eliottness, line 7:
```suggestion
// Unsupported target OS or architecture on a supported Go version
//            Unsupported OS        Unsupported Arch      Good Go Version
//go:build ((!linux && !darwin) || (!amd64 && !arm64)) && !go1.21
```

waf_unsupported_go.go
```diff
@@ -0,0 +1,16 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-present Datadog, Inc.
+
+// Supported OS/Arch but unsupported Go version
+//go:build (linux || darwin) && (amd64 || arm64) && go1.21
```
eliottness, line 7:
```suggestion
// Supported OS/Arch but unsupported Go version
//           Supported OS        Supported Arch     Bad Go Version
//go:build (linux || darwin) && (amd64 || arm64) && go1.21
```

eliottness (approved):

PR #20, opened 2023-07-18
ci: native go1.21 tests

Julio-Guerra:
Include more go1.21rc tests, as I figured out how to get the release candidate in the native jobs that were using `@actions/setup-go`.


eliottness (approved):
<h1>:+1:</h1>

PR #26, opened 2023-09-20
fix: handle.NewHandle() now accepts map instead of any

RomainMuller:
This is to avoid users mistakenly passing a JSON string instead of the parsed object.


Julio-Guerra (approved):

eliottness:
Actually we cannot do this because we ingest the unexported struct [rulesFragment](https://github.com/DataDog/dd-trace-go/blob/main/internal/appsec/rules_manager.go#L30). I will revert it until we either promote this struct as object model or find a workaround

PR #27, opened 2023-09-28
rename correctly allocator to cgoRefPool

eliottness:
Fix README errors and add precisions

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ The WAF bindings have multiple moving parts that are necessary to understand:
 - Handle: a object wrapper over the pointer to the C WAF Handle
 - Context: a object wrapper over a pointer to the C WAF Context
 - Encoder: whose goal is to construct a tree of Waf Objects to send to the WAF
-- CGORefPol: Does all allocation operations for the construction of Waf Objects and keep track of the equivalent go pointers
+- CGORefPool: Does all allocation operations for the construction of Waf Objects and keeps track of the equivalent go pointers
 - Decoder: Transforms Waf Objects returned from the WAF to usual go objects (e.g. maps, arrays, ...)
 - Library: The low-level API of bindings with the C world, made to provide Go type checking to it
 
@@ -86,14 +86,14 @@ union equivalent in go are interfaces and they are not compatible with C unions.
 that the Go `wafObject` struct has the same layout as the C one is to only use primitive types. So the only way to
 store a raw pointer is to use the `uintptr` type. But since `uintptr` do not have pointer semantics (and are just
 basically integers), we need another method to store the value as Go pointer because the GC will delete our data if it
-not reference bu Go pointers.
+is not referenced by Go pointers.
 
-That's where the `cgoRefPool` object comes into play: all new `wafObject` elements are created via this API whose especially
+That's where the `cgoRefPool` object comes into play: all new `wafObject` elements are created via this API which is especially
 built to make sure there is no gap for the Garbage Collector to exploit. From there, since underlying values of the
 `wafObject` are either arrays of wafObjects (for maps, structs and arrays) or string (for all ints, booleans and strings),
 we can store 2 slices of arrays and use `runtime.KeepAlive` in each code path to protect them from the GC.
 
-All these objects stored in the reference pool need to live throughout the use of the Waf Context associated.
+All these objects stored in the reference pool need to live throughout the use of the associated Waf Context.
 
 ### Typical call to Run()
 
@@ -101,30 +101,30 @@ Here is an example of the flow of operations on a simple call to Run():
 
 - Encode input data into Waf Objects and store references in the temporary pool
 - Lock the context mutex until the end of the call
-- store references from the temporary pool into the context level pool
+- Store references from the temporary pool into the context level pool
 - Call `ddwaf_run`
 - Decode the matches and actions
 
 ### CGO-less C Bindings
 
 The main component used to build C bindings without using CGO is called [purego](https://github.com/ebitengine/purego). The flow of execution on our side
-is to embed the C shared library using `go:embed`. Then to dump it into a file, load it using `dlopen` and to load the
-symbols using `dlsym`. And finally to call them.
+is to embed the C shared library using `go:embed`, dump it into a file, open the library using `dlopen`, load the
+symbols using `dlsym`, and finally call them.
 
 ⚠️ Keep in mind that **purego only works on linux/darwin for amd64/arm64 and so does go-libddwaf.**
 
 Another requirement of `libddwaf` is to have a FHS filesystem on your machine and, for linux, to provide `libc.so.6`,
-`libpthread.so.0` and `libm.so.6`, `libdl.so.2` as dynamic libraries.
+`libpthread.so.0`, `libm.so.6` and `libdl.so.2` as dynamic libraries.
 
 ## Contributing pitfalls
 
 - Cannot dlopen twice in the app lifetime on OSX. It messes with Thread Local Storage and usually finishes with a `std::bad_alloc()`
 - `keepAlive()` calls are here to prevent the GC from destroying objects too early
-- Since there is a stack switch between the go code and the C code, usually the only C stacktrace you will ever get is from GDB
+- Since there is a stack switch between the Go code and the C code, usually the only C stacktrace you will ever get is from GDB
 - If a segfault happens during a call to the C code, the goroutine stacktrace which has done the call is the one annotated with `[syscall]`
 - [GoLand](https://www.jetbrains.com/go/) does not support `CGO_ENABLED=0` (as of June 2023)
-- Keep in mind that we fully escape the type system. If you send the wrong data it will segfaults in the best cases but not always !
-- The structs in `ctypes.go` are here to reproduce the memory layout of the structs in `include/ddwaf.h` because pointer to these structs will be passed directly
+- Keep in mind that we fully escape the type system. If you send the wrong data it will segfault in the best cases but not always !
+- The structs in `ctypes.go` are here to reproduce the memory layout of the structs in `include/ddwaf.h` because pointers to these structs will be passed directly
 - Do not use `uintptr` as function arguments or results types, coming from `unsafe.Pointer` casts of Go values, because they escape the pointer analysis which can create wrongly optimized code and crash. Pointer arithmetic is of course necessary in such a library but must be kept in the same function scope.
 - GDB is available on arm64 but is not officially supported so it usually crashes pretty fast
-- No pointer from variables on the stack shall be sent to the C code because Go stacks can be moved during the C call. More on this [here](https://medium.com/@trinad536/escape-analysis-in-golang-fc81b78f3550)
+- No pointer to variables on the stack shall be sent to the C code because Go stacks can be moved during the C call. More on this [here](https://medium.com/@trinad536/escape-analysis-in-golang-fc81b78f3550)
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # go-libddwaf
 
 This project's goal is to produce a higher level API for the go bindings to [libddwaf](https://github.com/DataDog/libddwaf): DataDog in-app WAF.
-It consists of 2 separate entities: the bindings for the calls to libddwaf, and the encoder whose job is to convert _any_ go value to its libddwaf object representation.
+It consists of 2 separate entities: the bindings for the calls to libddwaf, and the encoder which job is to convert _any_ go value to its libddwaf object representation.
 
 An example usage would be:
 
@@ -47,10 +47,10 @@ The WAF bindings have multiple moving parts that are necessary to understand:
 
 - Handle: a object wrapper over the pointer to the C WAF Handle
 - Context: a object wrapper over a pointer to the C WAF Context
-- Encoder: whose goal is to construct a tree of Waf Objects to send to the WAF
+- Encoder: its goal is to construct a tree of Waf Objects to send to the WAF
 - CGORefPool: Does all allocation operations for the construction of Waf Objects and keeps track of the equivalent go pointers
 - Decoder: Transforms Waf Objects returned from the WAF to usual go objects (e.g. maps, arrays, ...)
-- Library: The low-level API of bindings with the C world, made to provide Go type checking to it
+- Library: The low-level go bindings to the C library, providing improved typing
 
 ```mermaid
 flowchart LR
@@ -123,8 +123,8 @@ Another requirement of `libddwaf` is to have a FHS filesystem on your machine an
 - Since there is a stack switch between the Go code and the C code, usually the only C stacktrace you will ever get is from GDB
 - If a segfault happens during a call to the C code, the goroutine stacktrace which has done the call is the one annotated with `[syscall]`
 - [GoLand](https://www.jetbrains.com/go/) does not support `CGO_ENABLED=0` (as of June 2023)
-- Keep in mind that we fully escape the type system. If you send the wrong data it will segfault in the best cases but not always !
+- Keep in mind that we fully escape the type system. If you send the wrong data it will segfault in the best cases but not always!
 - The structs in `ctypes.go` are here to reproduce the memory layout of the structs in `include/ddwaf.h` because pointers to these structs will be passed directly
 - Do not use `uintptr` as function arguments or results types, coming from `unsafe.Pointer` casts of Go values, because they escape the pointer analysis which can create wrongly optimized code and crash. Pointer arithmetic is of course necessary in such a library but must be kept in the same function scope.
-- GDB is available on arm64 but is not officially supported so it usually crashes pretty fast
+- GDB is available on arm64 but is not officially supported so it usually crashes pretty fast (as of June 2023)
 - No pointer to variables on the stack shall be sent to the C code because Go stacks can be moved during the C call. More on this [here](https://medium.com/@trinad536/escape-analysis-in-golang-fc81b78f3550)
--- a/README.md
+++ b/README.md
@@ -48,9 +48,9 @@ The WAF bindings have multiple moving parts that are necessary to understand:
 - Handle: a object wrapper over the pointer to the C WAF Handle
 - Context: a object wrapper over a pointer to the C WAF Context
 - Encoder: whose goal is to construct a tree of Waf Objects to send to the WAF
-- Allocator: Does all writing and allocation operations for the construction of Waf Objects
+- CGORefPol: Does all allocation operations for the construction of Waf Objects and keep track of the equivalent go pointers
 - Decoder: Transforms Waf Objects returned from the WAF to usual go objects (e.g. maps, arrays, ...)
-- Library: The library which wraps all calls to C code
+- Library: The low-level API of bindings with the C world, made to provide Go type checking to it
 
 ```mermaid
 flowchart LR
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -66,55 +66,65 @@ flowchart LR
...
+basically integers), we need another method to store the value as Go pointer because the GC will delete our data if it
+not reference bu Go pointers.
+
+That's where the `cgoRefPool` object comes into play: all new `wafObject` elements are created via this API whose especially
 built to make sure there is no gap for the Garbage Collector to exploit. From there, since underlying values of the
-`wafObject` are either arrays (for maps, structs and arrays) or string (for all ints, booleans and strings),
+`wafObject` are either arrays of wafObjects (for maps, structs and arrays) or string (for all ints, booleans and strings),
 we can store 2 slices of arrays and use `runtime.KeepAlive` in each code path to protect them from the GC.
 
+All these objects stored in the reference pool need to live throughout the use of the Waf Context associated.
```
Hellzy, line 96:
```suggestion
All these objects stored in the reference pool need to live throughout the use of the associated Waf Context.
```

README.md
```diff
@@ -48,9 +48,9 @@ The WAF bindings have multiple moving parts that are necessary to understand:
 - Handle: a object wrapper over the pointer to the C WAF Handle
 - Context: a object wrapper over a pointer to the C WAF Context
 - Encoder: whose goal is to construct a tree of Waf Objects to send to the WAF
-- Allocator: Does all writing and allocation operations for the construction of Waf Objects
+- CGORefPol: Does all allocation operations for the construction of Waf Objects and keep track of the equivalent go pointers
```
RomainMuller:
```suggestion
- CGORefPol: Does all allocation operations for the construction of WAF Objects and keeps track of the equivalent go pointers
```

README.md
```diff
@@ -66,55 +66,65 @@ flowchart LR
...
+All these objects stored in the reference pool need to live throughout the use of the Waf Context associated.
+
 ### Typical call to Run()
 
 Here is an example of the flow of operations on a simple call to Run():
 
-- Encode input data into Waf Objects
+- Encode input data into Waf Objects and store references in the temporary pool
 - Lock the context mutex until the end of the call
+- store references from the temporary pool into the context level pool
```
Hellzy, line 104:
```suggestion
- Store references from the temporary pool into the context level pool
```

README.md
```diff
@@ -48,9 +48,9 @@ The WAF bindings have multiple moving parts that are necessary to understand:
 - Handle: a object wrapper over the pointer to the C WAF Handle
 - Context: a object wrapper over a pointer to the C WAF Context
 - Encoder: whose goal is to construct a tree of Waf Objects to send to the WAF
-- Allocator: Does all writing and allocation operations for the construction of Waf Objects
+- CGORefPol: Does all allocation operations for the construction of Waf Objects and keep track of the equivalent go pointers
 - Decoder: Transforms Waf Objects returned from the WAF to usual go objects (e.g. maps, arrays, ...)
-- Library: The library which wraps all calls to C code
+- Library: The low-level API of bindings with the C world, made to provide Go type checking to it
```
RomainMuller, line 53:
```suggestion
- Library: The low-level go bindings to the C library, providing improved typings to go consumers
```

README.md
```diff
@@ -66,55 +66,65 @@ flowchart LR
...
 - Lock the context mutex until the end of the call
+- store references from the temporary pool into the context level pool
 - Call `ddwaf_run`
 - Decode the matches and actions
 
 ### CGO-less C Bindings
 
-The main component used to build C bindings without using CGO is called [purego](https://github.com/ebitengine/purego). The flow of execution on our side is to embed the C shared library using `go:embed`. Then to dump it into a file, load it using `dlopen` and to load the symbols using `dlsym`. And finally to call them.
+The main component used to build C bindings without using CGO is called [purego](https://github.com/ebitengine/purego). The flow of execution on our side
+is to embed the C shared library using `go:embed`. Then to dump it into a file, load it using `dlopen` and to load the
```
Hellzy, line 111: I think a simple comma separated sentence would be smoother to read "embed [...], dump it [...], load it [...] and load the symbols [...], and finally call them". Something like that, wdyt?

README.md
```diff
@@ -66,55 +66,65 @@ flowchart LR
...
     classDef hidden display: none;
 ```
 
-### Allocator
+### CGO Reference Pool
 
-The cgoRefPool is a pure Go cgoRefPool of `ddwaf_object` C values on the Go memory heap.
-the `cgoRefPool` go type is a way to make sure we can safely send go allocated data to the C side of the WAF
+The cgoRefPool type is a pure Go pointer pool of `ddwaf_object` C values on the Go memory heap.
+the `cgoRefPool` go type is a way to make sure we can safely send Go allocated data to the C side of the WAF
```
RomainMuller, line 83: Sometimes it's "go type", sometimes just type... I'd recommend making all "type" implicitly be go and to actually expressly specify when you're referring to a C type...

There is also a lack of punctuation at the end of this sentence...

README.md
```diff
@@ -66,55 +66,65 @@ flowchart LR
...
-The main component used to build C bindings without using CGO is called [purego](https://github.com/ebitengine/purego). The flow of execution on our side is to embed the C shared library using `go:embed`. Then to dump it into a file, load it using `dlopen` and to load the symbols using `dlsym`. And finally to call them.
+The main component used to build C bindings without using CGO is called [purego](https://github.com/ebitengine/purego). The flow of execution on our side
+is to embed the C shared library using `go:embed`. Then to dump it into a file, load it using `dlopen` and to load the
+symbols using `dlsym`. And finally to call them.
 
 ⚠️ Keep in mind that **purego only works on linux/darwin for amd64/arm64 and so does go-libddwaf.**
 
-Another requirement of `libddwaf` is to have a FHS filesystem on your machine and, for linux, to provide `libc.so.6`, `libpthread.so.0` and `libm.so.6`, `libdl.so.2` as dynamic libraries.
+Another requirement of `libddwaf` is to have a FHS filesystem on your machine and, for linux, to provide `libc.so.6`,
+`libpthread.so.0` and `libm.so.6`, `libdl.so.2` as dynamic libraries.
```
Hellzy, line 117:
```suggestion
`libpthread.so.0`, `libm.so.6` and `libdl.so.2` as dynamic libraries.
```

README.md
```diff
@@ -66,55 +66,65 @@ flowchart LR
...
-## Contributing usual pitfalls
+## Contributing pitfalls
 
-- Cannot dlopen twice in the app lifetime on OSX
-- `runtime.KeepAlive()` calls are here to prevent the GC from destroying objects too early
-- Since there is a stack switch between the go code and the C code, usually the only C stacktrace you will ever get is from gdb
-- If a segfault happens during a call to the C code, the goroutine stacktrace which has done the call is the one annotated with `[syscall]`.
+- Cannot dlopen twice in the app lifetime on OSX. It messes with Thread Local Storage and usually finishes with a `std::bad_alloc()`
+- `keepAlive()` calls are here to prevent the GC from destroying objects too early
+- Since there is a stack switch between the go code and the C code, usually the only C stacktrace you will ever get is from GDB
```
Hellzy, line 123:
```suggestion
- Since there is a stack switch between the Go code and the C code, usually the only C stacktrace you will ever get is from GDB
```

README.md
```diff
@@ -66,55 +66,65 @@ flowchart LR
...
+- If a segfault happens during a call to the C code, the goroutine stacktrace which has done the call is the one annotated with `[syscall]`
 - [GoLand](https://www.jetbrains.com/go/) does not support `CGO_ENABLED=0` (as of June 2023)
-- Keep in mind that we fully escape the type system. If you send the wrong data it will segfaults in the best cases but not always!
-- The structs in `ctypes.go` are here to reproduce the memory layout of the structs in `include/ddwaf.h` because pointer to these structs will be passed directly.
+- Keep in mind that we fully escape the type system. If you send the wrong data it will segfaults in the best cases but not always !
```
Hellzy, line 126:
```suggestion
- Keep in mind that we fully escape the type system. If you send the wrong data it will segfault in the best cases but not always !
```

README.md
```diff
@@ -66,55 +66,65 @@ flowchart LR
...
+The cgoRefPool type is a pure Go pointer pool of `ddwaf_object` C values on the Go memory heap.
+the `cgoRefPool` go type is a way to make sure we can safely send Go allocated data to the C side of the WAF
 The main issue is the following: the `wafObject` uses a C union to store the tree structure of the full object,
 union equivalent in go are interfaces and they are not compatible with C unions. The only way to be 100% sure
 that the Go `wafObject` struct has the same layout as the C one is to only use primitive types. So the only way to
 store a raw pointer is to use the `uintptr` type. But since `uintptr` do not have pointer semantics (and are just
-basically integers), we need another structure to store the value as Go pointer because the GC is lurking. That's
-where the `cgoRefPool` object comes into play: all new `wafObject` elements are created via this API whose especially
+basically integers), we need another method to store the value as Go pointer because the GC will delete our data if it
+not reference bu Go pointers.
```
RomainMuller: I find this isn't all too clear... Not sure how to rephrase... it also feels to me like some diagrams could help understand the principle of how it works.

I'd also avoir using `method` here because this is a programming term with a clear meaning... `approach` might be better.

README.md
```diff
@@ -66,55 +66,65 @@ flowchart LR
...
+- Cannot dlopen twice in the app lifetime on OSX. It messes with Thread Local Storage and usually finishes with a `std::bad_alloc()`
+- `keepAlive()` calls are here to prevent the GC from destroying objects too early
+- Since there is a stack switch between the go code and the C code, usually the only C stacktrace you will ever get is from GDB
+- If a segfault happens during a call to the C code, the goroutine stacktrace which has done the call is the one annotated with `[syscall]`
 - [GoLand](https://www.jetbrains.com/go/) does not support `CGO_ENABLED=0` (as of June 2023)
-- Keep in mind that we fully escape the type system. If you send the wrong data it will segfaults in the best cases but not always!
-- The structs in `ctypes.go` are here to reproduce the memory layout of the structs in `include/ddwaf.h` because pointer to these structs will be passed directly.
+- Keep in mind that we fully escape the type system. If you send the wrong data it will segfaults in the best cases but not always !
+- The structs in `ctypes.go` are here to reproduce the memory layout of the structs in `include/ddwaf.h` because pointer to these structs will be passed directly
```
Hellzy, line 127:
```suggestion
- The structs in `ctypes.go` are here to reproduce the memory layout of the structs in `include/ddwaf.h` because pointers to these structs will be passed directly
```

README.md
```diff
@@ -66,55 +66,65 @@ flowchart LR
...
 - Do not use `uintptr` as function arguments or results types, coming from `unsafe.Pointer` casts of Go values, because they escape the pointer analysis which can create wrongly optimized code and crash. Pointer arithmetic is of course necessary in such a library but must be kept in the same function scope.
+- GDB is available on arm64 but is not officially supported so it usually crashes pretty fast
+- No pointer from variables on the stack shall be sent to the C code because Go stacks can be moved during the C call. More on this [here](https://medium.com/@trinad536/escape-analysis-in-golang-fc81b78f3550)
```
Hellzy, line 130:
```suggestion
- No pointer to variables on the stack shall be sent to the C code because Go stacks can be moved during the C call. More on this [here](https://medium.com/@trinad536/escape-analysis-in-golang-fc81b78f3550)
```

README.md
```diff
@@ -66,55 +66,65 @@ flowchart LR
...
 The main issue is the following: the `wafObject` uses a C union to store the tree structure of the full object,
 union equivalent in go are interfaces and they are not compatible with C unions. The only way to be 100% sure
 that the Go `wafObject` struct has the same layout as the C one is to only use primitive types. So the only way to
 store a raw pointer is to use the `uintptr` type. But since `uintptr` do not have pointer semantics (and are just
-basically integers), we need another structure to store the value as Go pointer because the GC is lurking. That's
-where the `cgoRefPool` object comes into play: all new `wafObject` elements are created via this API whose especially
+basically integers), we need another method to store the value as Go pointer because the GC will delete our data if it
+not reference bu Go pointers.
+
+That's where the `cgoRefPool` object comes into play: all new `wafObject` elements are created via this API whose especially
```
RomainMuller:
```suggestion
That's where the `cgoRefPool` object comes into play: all new `wafObject` elements are created via this API which is especially
```

Hellzy (changes requested):

README.md
```diff
@@ -66,55 +66,65 @@ flowchart LR
...
 built to make sure there is no gap for the Garbage Collector to exploit. From there, since underlying values of the
-`wafObject` are either arrays (for maps, structs and arrays) or string (for all ints, booleans and strings),
+`wafObject` are either arrays of wafObjects (for maps, structs and arrays) or string (for all ints, booleans and strings),
```
RomainMuller, line 93: There is inconsistent backticking of type names...

README.md
RomainMuller: Consistently case "WAF"

```suggestion
All these objects stored in the reference pool need to live throughout the use of the WAF Context associated.
```

README.md
```diff
@@ -66,55 +66,65 @@ flowchart LR
...
 we can store 2 slices of arrays and use `runtime.KeepAlive` in each code path to protect them from the GC.
 
+All these objects stored in the reference pool need to live throughout the use of the Waf Context associated.
+
 ### Typical call to Run()
 
 Here is an example of the flow of operations on a simple call to Run():
 
-- Encode input data into Waf Objects
+- Encode input data into Waf Objects and store references in the temporary pool
```
RomainMuller, line 102:
```suggestion
- Encode input data into WAF Objects and store references in the temporary pool
```

README.md
```diff
@@ -66,55 +66,65 @@ flowchart LR
...
 - Lock the context mutex until the end of the call
```
RomainMuller, line 103: Not pictured here: unlocking the mutex

README.md
```diff
@@ -66,55 +66,65 @@ flowchart LR
...
+- store references from the temporary pool into the context level pool
 - Call `ddwaf_run`
 - Decode the matches and actions
 
 ### CGO-less C Bindings
 
-The main component used to build C bindings without using CGO is called [purego](https://github.com/ebitengine/purego). The flow of execution on our side is to embed the C shared library using `go:embed`. Then to dump it into a file, load it using `dlopen` and to load the symbols using `dlsym`. And finally to call them.
+The main component used to build C bindings without using CGO is called [purego](https://github.com/ebitengine/purego). The flow of execution on our side
```
RomainMuller, line 110: I find `build` a bit overloaded here especially as there are mentions of `cgo` here... I also find a more clinical phrasing is clearer (feel free to disagree):

```suggestion
This library uses [purego](https://github.com/ebitengine/purego) to implement C bindings without requiring use of CGO at compilation time. The high-level workflow
```

README.md
RomainMuller:
```suggestion
is to embed the C shared library using `go:embed`. The library is dumped into a temporary file and loaded `dlopen` then resolve the relevant
```

README.md
```diff
@@ -66,55 +66,65 @@ flowchart LR
...
+is to embed the C shared library using `go:embed`. Then to dump it into a file, load it using `dlopen` and to load the
+symbols using `dlsym`. And finally to call them.
```
RomainMuller:
```suggestion
symbols using `dlsym`.
```

README.md
```diff
@@ -66,55 +66,65 @@ flowchart LR
...
 
 ⚠️ Keep in mind that **purego only works on linux/darwin for amd64/arm64 and so does go-libddwaf.**
```
RomainMuller, line 114: I like to nest this kins of stuff in a blockauote for added viz

```suggestion
> :warning: Keep in mind that **purego only works on linux/darwin for amd64/arm64 and so does go-libddwaf.**
```

README.md
```diff
@@ -66,55 +66,65 @@ flowchart LR
...
 
-Another requirement of `libddwaf` is to have a FHS filesystem on your machine and, for linux, to provide `libc.so.6`, `libpthread.so.0` and `libm.so.6`, `libdl.so.2` as dynamic libraries.
+Another requirement of `libddwaf` is to have a FHS filesystem on your machine and, for linux, to provide `libc.so.6`,
```
RomainMuller, line 116: This starts with "another requirement" but I dont see other requirements mentioned here in the diff... this is a bit surprising.

It's also not clear why a FHS layout is necessary... What happens if you don't?
I don't think FHS is actually required but certain files have to be in certain places.

I'd call this out only as a linux requirement since macOS FS layout is standardized and 100% controlled by apple.

README.md
eliottness, line 89:
```suggestion
basically integers), we need another method to store the value as Go pointer because the GC will delete our data if it
is not referenced by Go pointers.
```

README.md
```diff
@@ -66,55 +66,65 @@ flowchart LR
...
-the `cgoRefPool` go type is a way to make sure we can safely send go allocated data to the C side of the WAF
+The cgoRefPool type is a pure Go pointer pool of `ddwaf_object` C values on the Go memory heap.
+the `cgoRefPool` go type is a way to make sure we can safely send Go allocated data to the C side of the WAF
 The main issue is the following: the `wafObject` uses a C union to store the tree structure of the full object,
 union equivalent in go are interfaces and they are not compatible with C unions. The only way to be 100% sure
 that the Go `wafObject` struct has the same layout as the C one is to only use primitive types. So the only way to
 store a raw pointer is to use the `uintptr` type. But since `uintptr` do not have pointer semantics (and are just
-basically integers), we need another structure to store the value as Go pointer because the GC is lurking. That's
-where the `cgoRefPool` object comes into play: all new `wafObject` elements are created via this API whose especially
+basically integers), we need another method to store the value as Go pointer because the GC will delete our data if it
```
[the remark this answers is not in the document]
eliottness: https://github.com/DataDog/go-libddwaf/pull/27/files#r1341057410

README.md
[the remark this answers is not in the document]
eliottness: Both are working grammar-wise but fine

README.md
RomainMuller: The `.so` files are necessarily dynamic libraries, aren't they? It might be interesting to specify how to ensure they're installed on the major distributions (if any special action is needed).

README.md
eliottness, line 112:
```suggestion
is to embed the C shared library using `go:embed`, dump it into a file, open the library using `dlopen`, load the
symbols using `dlsym`, and finally call them.
```

README.md
```diff
@@ -66,55 +66,65 @@ flowchart LR
...
+`libpthread.so.0` and `libm.so.6`, `libdl.so.2` as dynamic libraries.
 
-## Contributing usual pitfalls
+## Contributing pitfalls
 
-- Cannot dlopen twice in the app lifetime on OSX
-- `runtime.KeepAlive()` calls are here to prevent the GC from destroying objects too early
-- Since there is a stack switch between the go code and the C code, usually the only C stacktrace you will ever get is from gdb
-- If a segfault happens during a call to the C code, the goroutine stacktrace which has done the call is the one annotated with `[syscall]`.
+- Cannot dlopen twice in the app lifetime on OSX. It messes with Thread Local Storage and usually finishes with a `std::bad_alloc()`
```
RomainMuller, line 121: Should specify this is for a given shared library... The `purego` doc says if you dlopen the same object multiple times it reuses the handles instead of reloading... considered this could be a bug on their end?

README.md
eliottness: https://github.com/DataDog/go-libddwaf/pull/27/files#r1341060938

README.md
```diff
@@ -66,55 +66,65 @@ flowchart LR
...
+- `keepAlive()` calls are here to prevent the GC from destroying objects too early
```
RomainMuller, line 122: I'd phrase this slightly differently -- this is to keep memory that is sent over to C-land where it's no longer managed by the go runtime

README.md
RomainMuller: What does this mean concretely? Means the C code will always be invisible unless you have a gdb attached, and C frames won't unwind all the way into go? TBH this makes me think `CONTRIBUTING.md` should give information on how to debug your code and all...

README.md
```diff
@@ -66,55 +66,65 @@ flowchart LR
...
+- Since there is a stack switch between the go code and the C code, usually the only C stacktrace you will ever get is from GDB
+- If a segfault happens during a call to the C code, the goroutine stacktrace which has done the call is the one annotated with `[syscall]`
```
RomainMuller, line 124: And example is worth a thousand words

README.md
RomainMuller: No space before ! in english

```suggestion
- Keep in mind that we fully escape the type system. If you send the wrong data it will SEGFAULT in the best cases but not always!
```

This makes me wonder if there are resources about C FFI / C bridging somewhere because these all are common pitfalls of doing language hops... and there are probably common and well documented techniques to deal with these...

README.md
RomainMuller:
```suggestion
- The structs in `ctypes.go` mirror the memory layout of those in `include/ddwaf.h` so that pointers to these structs can be passed directly
```

README.md
```diff
@@ -66,55 +66,65 @@ flowchart LR
...
 - [GoLand](https://www.jetbrains.com/go/) does not support `CGO_ENABLED=0` (as of June 2023)
-- Keep in mind that we fully escape the type system. If you send the wrong data it will segfaults in the best cases but not always!
-- The structs in `ctypes.go` are here to reproduce the memory layout of the structs in `include/ddwaf.h` because pointer to these structs will be passed directly.
+- Keep in mind that we fully escape the type system. If you send the wrong data it will segfaults in the best cases but not always !
+- The structs in `ctypes.go` are here to reproduce the memory layout of the structs in `include/ddwaf.h` because pointer to these structs will be passed directly
 - Do not use `uintptr` as function arguments or results types, coming from `unsafe.Pointer` casts of Go values, because they escape the pointer analysis which can create wrongly optimized code and crash. Pointer arithmetic is of course necessary in such a library but must be kept in the same function scope.
+- GDB is available on arm64 but is not officially supported so it usually crashes pretty fast
```
RomainMuller, line 129: Is that still true? I thought it was now a tier 1 platform in LLVM?

README.md
RomainMuller: Worth explaining how to notice this at code level?

RomainMuller (changes requested):

README.md
eliottness:
```suggestion
The cgoRefPool type is a pure Go pointer pool of `ddwaf_object` C values on the Go memory heap.
It is a way to make sure we can safely send Go allocated data to the C side of the WAF without the Garbage Collector reclaiming the memory too early.
```

Do not hesitate to write suggestions in case like this, it's easier for everyone :+1:

README.md
eliottness:
```suggestion
`wafObject` are either arrays of `wafObject` (for maps, structs and arrays) or string (for all ints, booleans and strings),
```

README.md
eliottness: In my  mind the part about "until the end of the call" is like a defer statement. I think this is clear enough because you inferred this part yourself and you literrally are the targeted reeader.

README.md
eliottness: sure why not

README.md
eliottness: It stemmed in my mind of the fact that we sometimes need the full path of the libraries, sometimes we don't and the rules are not really straightforward. At this point T guess this is obvious anyway so I am good to remove it

```suggestion
Another requirement of `libddwaf` on linux is to provide `libc.so.6`,
```

README.md
eliottness: Purego doc says this but on their side they have no state of this. They only rely on the POSIX doc of it. But I am sure of it because I spent weeks on this shit. This is in fact a bug (or a feature) of the way OSX designed this part of their linker. More specifically they do not support the following actions: dlopen(/tmp/toto.dylib) -> dlclose() -> mv /tmp/toto.dylib /tmp/tata.dylib -> dlopen(/tmp/tata.dylib) -> std::bad_alloc()

README.md
eliottness: I can't built a pleasant sentence. Do you have a suggestion ?

README.md
eliottness: Do you have a example in mind ?

README.md
eliottness: Compiled it myself and still. Maye this was linked somehow to VMware I don't know. But still we can add something like this
```suggestion
- GDB is available on arm64 but is not officially supported so it usually crashes pretty fast (as of June 2023
```

README.md
RomainMuller: I'm saying these make it easier to code against the C code in go... consumers can be internal, too.

README.md
RomainMuller: I do when I have a clear & confident idea of what I'd like to see... in this case I felt like this was very opinion land and I was also not sure about the rest of the doc just yet so it could also have been a bad idea

README.md
RomainMuller: Nah it's fine. I'm a bit surprised by also... that's also mildly unsurprising

README.md
eliottness:
```suggestion
- Library: The low-level go bindings to the C library, providing improved typing
```

README.md
eliottness: This is done in the article I linked

RomainMuller (approved):

PR #29, opened 2023-10-11
chore: remove avoidable use of unsafe

RomainMuller:
Use generics and a tiny interface to remove a perfectly avoidable use of reflection+unsafe to provision a handle, and remove a possible mis-use opportunity.


eliottness (approved):
LGTM (Thanks for deps bump) 👍

PR #30, opened 2023-10-12
Update WAF to 13.0

Hellzy:
## Upgrade WAF to 13.0
- Regroup WAF outputs in `Result`
- Handle actions as WAF objects
- Handle WAF result as WAF objects
- Handle WAF diagnostics
- Handle new `derivatives` output (API sec schemas)
- Handle float/int/uint/bool encoding/decoding (used to be strings)
- Drop `RulesetInfo` and associated helper bindings


context.go
```diff
@@ -58,7 +58,7 @@ func NewContext(handle *Handle) *Context {
 // timeout value. It returns the matches as a JSON string (usually opaquely used) along with the corresponding
 // actions in any. In case of an error, matches and actions can still be returned, for instance in the case of a
 // timeout error. Errors can be tested against the RunError type.
-func (context *Context) Run(addressesToData map[string]any, timeout time.Duration) (matches []byte, actions []string, err error) {
+func (context *Context) Run(addressesToData map[string]any, timeout time.Duration) (events []interface{}, actions []string, err error) {
```
RomainMuller:
```suggestion
func (context *Context) Run(addressesToData map[string]any, timeout time.Duration) (events []any, actions []string, err error) {
```

context.go
```diff
@@ -90,7 +90,7 @@ func (context *Context) Run(addressesToData map[string]any, timeout time.Duratio
 	return context.run(obj, timeout, &encoder.cgoRefs)
 }
 
-func (context *Context) run(obj *wafObject, timeout time.Duration, cgoRefs *cgoRefPool) ([]byte, []string, error) {
+func (context *Context) run(obj *wafObject, timeout time.Duration, cgoRefs *cgoRefPool) ([]interface{}, []string, error) {
```
RomainMuller:
```suggestion
func (context *Context) run(obj *wafObject, timeout time.Duration, cgoRefs *cgoRefPool) ([]any, []string, error) {
```

context.go
```diff
@@ -102,15 +102,15 @@ func (context *Context) run(obj *wafObject, timeout time.Duration, cgoRefs *cgoR
...
 	if err == ErrTimeout {
 		context.timeoutCount.Inc()
 	}
 
-	return matches, actions, err
+	return events, actions, err
 }
 
-func unwrapWafResult(ret wafReturnCode, result *wafResult) (matches []byte, actions []string, err error) {
+func unwrapWafResult(ret wafReturnCode, result *wafResult) (matches []interface{}, actions []string, err error) {
```
RomainMuller:
```suggestion
func unwrapWafResult(ret wafReturnCode, result *wafResult) (matches []any, actions []string, err error) {
```

decoder.go
```diff
@@ -57,17 +57,53 @@ func decodeRuleIdArray(obj *wafObject) ([]string, error) {
 	return ruleIds, nil
 }
 
-func decodeActions(cActions uintptr, size uint64) []string {
-	if size == 0 {
-		return nil
+func decodeObject(obj *wafObject) (interface{}, error) {
```
RomainMuller, line 60:
```suggestion
func decodeObject(obj *wafObject) (any, error) {
```

decoder.go
```diff
@@ -57,17 +57,53 @@ func decodeRuleIdArray(obj *wafObject) ([]string, error) {
...
+	case wafArrayType:
+		return decodeArray(obj)
+	case wafStringType:
+		return gostringSized(cast[byte](obj.value), obj.nbEntries), nil
+	default:
+		return nil, errUnsupportedValue
+	}
+}
+
+func decodeArray(obj *wafObject) ([]interface{}, error) {
```
RomainMuller, line 73:
```suggestion
func decodeArray(obj *wafObject) ([]any, error) {
```

decoder.go
```diff
@@ -57,17 +57,53 @@ func decodeRuleIdArray(obj *wafObject) ([]string, error) {
...
+	if obj._type != wafArrayType {
+		return nil, errInvalidObjectType
+	}
+
+	var events []interface{}
```
RomainMuller, line 78: In this case we know how much capacity is required upfront, so if we `make` the map we might safe some reallocs.

```suggestion
	events := make([]any, obj.nbEntries)
```

decoder.go
```diff
@@ -57,17 +57,53 @@ func decodeRuleIdArray(obj *wafObject) ([]string, error) {
...
+		if err != nil {
+			return nil, err
+		}
+		events = append(events, val)
+	}
+
+	return events, nil
+}
+
+func decodeMap(obj *wafObject) (map[string]interface{}, error) {
```
RomainMuller, line 92:
```suggestion
func decodeMap(obj *wafObject) (map[string]any, error) {
```

decoder.go
```diff
@@ -57,17 +57,53 @@ func decodeRuleIdArray(obj *wafObject) ([]string, error) {
...
+	if obj._type != wafMapType {
+		return nil, errInvalidObjectType
 	}
 
-	actions := make([]string, size)
-	for i := uint64(0); i < size; i++ {
-		// This line does the following operation without casts:
-		// gostring(*(cActions + i * sizeof(ptr)))
-		actions[i] = gostring(*castWithOffset[*byte](cActions, i))
+	result := map[string]interface{}{}
```
RomainMuller, line 97:
```suggestion
	result := make(map[string]any, obj.nbEntries)
```

RomainMuller (commented):
I'm pretty partial to `any` instead of `interface{}`. Also it's best if we can pre-allocate capacity for slices & maps when we know the item count ahead of time (which in marshaling `wafObjects` is always the case)

context.go
```diff
@@ -58,7 +58,7 @@ func NewContext(handle *Handle) *Context {
 // timeout value. It returns the matches as a JSON string (usually opaquely used) along with the corresponding
 // actions in any. In case of an error, matches and actions can still be returned, for instance in the case of a
 // timeout error. Errors can be tested against the RunError type.
-func (context *Context) Run(addressesToData map[string]any, timeout time.Duration) (matches []byte, actions []string, err error) {
+func (context *Context) Run(addressesToData map[string]any, timeout time.Duration) (events []interface{}, actions []string, derivatives map[string]interface{}, err error) {
```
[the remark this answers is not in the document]
Hellzy: That's actually one of the things I wanted your opinion about. I'm in favour of this 👍

context.go
```diff
@@ -58,7 +58,7 @@ func NewContext(handle *Handle) *Context {
 // timeout value. It returns the matches as a JSON string (usually opaquely used) along with the corresponding
 // actions in any. In case of an error, matches and actions can still be returned, for instance in the case of a
 // timeout error. Errors can be tested against the RunError type.
-func (context *Context) Run(addressesToData map[string]any, timeout time.Duration) (matches []byte, actions []string, err error) {
+func (context *Context) Run(addressesToData map[string]any, timeout time.Duration) (res Result, err error) {
```
eliottness, line 61: The whole function does not need named return so we can remove them.

eliottness:
```suggestion
func (context *Context) Run(addressesToData map[string]any, timeout time.Duration) (Result, error) {
```

context.go
```diff
@@ -76,7 +76,7 @@ func (context *Context) Run(addressesToData map[string]any, timeout time.Duratio
 	}
 	obj, err := encoder.Encode(addressesToData)
 	if err != nil {
-		return nil, nil, err
+		return res, err
```
eliottness, line 79:
```suggestion
		return nil, err
```

context.go
```diff
@@ -102,36 +102,47 @@ func (context *Context) run(obj *wafObject, timeout time.Duration, cgoRefs *cgoR
...
-	return matches, actions, err
+	return Result{
+		Events:      events,
+		Actions:     actions,
+		Derivatives: derivatives,
+	}, err
 }
 
-func unwrapWafResult(ret wafReturnCode, result *wafResult) (matches []byte, actions []string, err error) {
+func unwrapWafResult(ret wafReturnCode, result *wafResult) (matches []interface{}, actions []string, derivatives map[string]interface{}, err error) {
```
eliottness, line 117: Can you choose once and for all between "matches" and "events"? The `Result` time is about events but here we talk about matches

eliottness: I would definitively make this function return `Result` also

context.go
```diff
@@ -102,36 +102,47 @@ func (context *Context) run(obj *wafObject, timeout time.Duration, cgoRefs *cgoR
 	ret := wafLib.wafRun(context.cContext, obj, result, uint64(timeout/time.Microsecond))
 
 	context.totalRuntimeNs.Add(result.total_runtime)
-	matches, actions, err := unwrapWafResult(ret, result)
+	events, actions, derivatives, err := unwrapWafResult(ret, result)
 	if err == ErrTimeout {
 		context.timeoutCount.Inc()
 	}
```
eliottness, line 108: You could definitively do a `defer` out of this and directly `return unwrapWafResult(...)` if `unwrapWafResult` were to return `Result`. Just a suggestion tho

waf_dl.go
```diff
@@ -159,6 +154,10 @@ func (waf *wafDl) wafResultFree(result *wafResult) {
 	keepAlive(result)
 }
 
+func (waf *wafDl) wafObjectFree(obj *wafObject) {
+	waf.syscall(waf.objectFree, ptrToUintptr(obj))
```
eliottness, line 158:
```suggestion
	waf.syscall(waf.objectFree, ptrToUintptr(obj))
	keepAlive(obj)
```

eliottness: pointer = keepAlive

eliottness (changes requested):
I'll let you change all the interface{} to any and fix the CI then I will take a second look at it. Almost ready

context.go
Hellzy: This means changing the return type to `*Result`. I don't think it's needed since `err` is what is used to know if `res` should be used. An argument could be made for copy saving since the result bounces around a couple functions, but the structure is small and will get shallow copied so it should be neglectable.

context.go
Hellzy: Named return type here allows default initialisation for early return cases of `res`. Another way to do it would be simply `var res Result` at the top of the func, which is less idiomatic.

context.go
Hellzy: aa96a06e42dcf737a57ce99253b77b3be1852fd3

context.go
Hellzy: Is there a benefit I'm not seeing? It doesn't make the code easier to read and doesn't seem to solve a problem either.
Looks like using `defer` for the sake of it.

decoder.go
Hellzy: [0e8cf0a](https://github.com/DataDog/go-libddwaf/pull/30/commits/0e8cf0ae29eee053563ea20a196700e98b9eb83f)

context.go
eliottness: I could argue about the happy path and some more stuff but as I said it's just a suggestion anyway. I am closing it

eliottness (approved):
Lgtm

PR #31, opened 2023-10-16
chore: refactor away `libDl` (merge in `wafDl`)

RomainMuller:
Removes unnecessary use of reflection when resolving the symbols from libddwaf, and merges the `libDl` struct directly within the `wafDl` struct.

Additionally implemented some quality-of-life improvements around handling of the embedded library object:
- the temporary file is now platform-appropriately named;
- the temporary file is closed by the `dumpWafLibrary` function, which only returns the file path, removing the need to close the file everywhere;
- the WAF dynamic library tests are now centralized on a single entry point that fans out to the relevant implementation depending on `runtime.GOARCH`.


eliottness (approved):
<h1>LGTM :+1:</h1>

PR #33, opened 2023-11-02
Diagnostics: export fields

Hellzy:
We need to be able to read those fields to actually report diagnostics.
We could either implement getters or simply export the fields. The second option seems more convenient and appropriate.


RomainMuller (approved):

Hellzy:
We'll include this in 2.0.

Hellzy:
Cherry-picked in https://github.com/DataDog/go-libddwaf/pull/37, closing

PR #37, opened 2023-11-02
Update WAF to 1.13.0 (v2.0.0)

Hellzy:
Preparing for the 2.0.0 release, this change re-introduces the update to WAF 1.13.0, and tweaks the `Diagnostics` struct to export its fields 
- Update WAF to 13.0 (see https://github.com/DataDog/go-libddwaf/pull/30)
- Export `Diagnostics` fields for external use


handle.go
```diff
@@ -64,38 +64,33 @@ func NewHandle(rules any, keyObfuscatorRegex string, valueObfuscatorRegex string
...
-
-	errorsMap, err := decodeErrors(&cRulesetInfo.errors)
-	if err != nil { // Something is very wrong
-		return nil, fmt.Errorf("could not decode the WAF ruleset errors: %w", err)
+	diags, err := decodeDiagnostics(diagnosticsWafObj)
+	if err != nil {
+		return nil, fmt.Errorf("could not decode the WAF diagnostics: %w", err)
 	}
 
+	defer wafLib.wafObjectFree(diagnosticsWafObj)
```
eliottness, line 82: From the ddwaf_object_free doc, it does not free the top level object. Do we need to free it ?

eliottness (approved):

handle.go
eliottness: Ok it's irrelevant, the top level object is owned by the go runtime

PR #38, opened 2023-11-02
Another copy bites the dust

eliottness:
Since we upgraded the WAF to a version higher than 1.11, the input passed in are all read-only on the WAF side. Since this is the case we can remove unnecessary copies on our side


cgo_ref_pool.go
```diff
@@ -30,12 +30,9 @@ func (refPool *cgoRefPool) append(newRefs cgoRefPool) {
 }
 
 func (refPool *cgoRefPool) AllocCString(str string) uintptr {
-	goArray := make([]byte, len(str)+1)
-	copy(goArray, str)
-	refPool.stringRefs = append(refPool.stringRefs, goArray)
-	goArray[len(str)] = 0 // Null termination byte for C strings
-
-	return sliceToUintptr(goArray)
+	nullTerminated := str + "\x00"
```
Julio-Guerra, line 33: This can lead to a copy whereas libddwaf can manage non-null terminated strings via `ddwaf_object_stringl()`.
So I suggest we avoid this concatenation, by for example removing the existence of `AllocCString` (which has an unclear name with the 0-copy approach) and revisit this API to make it possible to avoid the concatenation here 🙏

eliottness: Today `AllocCString` is not used for string WAF objects, it is only used to allocated strings for the obfuscator config. `AllocWafString` is used for all cases concerning the WAF objects

eliottness: Documented the function

Hellzy (approved):

PR #39, opened 2023-11-03
Update WAF to 1.15.0

RomainMuller:
Also re-wrote the `update.sh` to be in go, for improved expressiveness. This includes dropping the `strip` pass as the `musl` linux images are already stripped, and this was just a waste of time.


context.go
```diff
@@ -99,7 +118,7 @@ func (context *Context) run(obj *wafObject, timeout time.Duration, cgoRefs *cgoR
 	result := new(wafResult)
 	defer wafLib.wafResultFree(result)
 
-	ret := wafLib.wafRun(context.cContext, obj, result, uint64(timeout/time.Microsecond))
+	ret := wafLib.wafRun(context.cContext, persistentData, ephemeralData, result, uint64(timeout/time.Microsecond))
```
eliottness, line 121: You have to change `waf_dl_unsupported.go` to fix the error on the windows runners: `too many arguments in call to wafLib.wafRun`

context.go
```diff
@@ -54,12 +54,26 @@ func NewContext(handle *Handle) *Context {
...
+}
+
+// Run encodes the given persistentAddressToData and ephemeralAddressToData values and runs them against the WAF rules
+// within the given timeout value. It returns the matches as a JSON string (usually opaquely used) along with the
+// corresponding actions in any. In case of an error, matches and actions can still be returned, for instance in the
+// case of a timeout error. Errors can be tested against the RunError type.
+func (context *Context) Run(addressData RunAddressData, timeout time.Duration) (res Result, err error) {
+	if addressData.isEmpty() {
 		return
 	}
```
eliottness, line 79: Let's add something to make sure both `addressData.persistent` and `addressData.ephemeral` cannot be nil at the same time

context.go
```diff
@@ -74,7 +88,12 @@ func (context *Context) Run(addressesToData map[string]any, timeout time.Duratio
 		containerMaxSize: wafMaxContainerSize,
 		objectMaxDepth:   wafMaxContainerDepth,
 	}
-	obj, err := encoder.Encode(addressesToData)
+	persistentData, err := encoder.Encode(addressData.persistent)
+	if err != nil {
+		return res, err
+	}
+
+	ephemeralData, err := encoder.Encode(addressData.ephemeral)
```
eliottness, line 96: You will have to find a way to bypass the usual behaviour of `Encode` because in case the input is nil it will return a `errUnsupportedValue`.

eliottness (commented):
First look at it, a couple of nits:
* Can you add tests that uses permanent and ephemeral addresses at the same time?
* Can you put the right permissions back on the embedded files ? (755 instead of 644)
* The embedded libddwaf on OSX where named with `_` before for a very specific reason but I can't reproduce it since I don't have a Mac anymore, let's keep it in the back of our mind for now.
* As you can see in the `run()` function, we keep track of the ddwaf_object parameter in the context struct because it is required for them to live all the way to the end of the context. My guess is that this is not the case for ephemeral addresses. Can you take a look at this with @Anilm3 ?

embed_darwin_arm64.go
```diff
@@ -9,7 +9,7 @@ package waf
 
 import _ "embed" // Needed for go:embed
 
-//go:embed lib/darwin-arm64/_libddwaf.dylib
+//go:embed lib/darwin-arm64/libddwaf.dylib
```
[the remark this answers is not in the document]
eliottness: I can't recall exactly but before all the files (even on linux) were named this way because there was a conflit with the static libraries. But I kept the ones of OSX because the linker was doing garbage-tier errors, like interpreting this line as a dependency to load which was reaaaaally wierd (as always). We can try to check if this is happening again

context.go
```diff
@@ -54,12 +54,26 @@ func NewContext(handle *Handle) *Context {
...
-// timeout value. It returns the matches as a JSON string (usually opaquely used) along with the corresponding
-// actions in any. In case of an error, matches and actions can still be returned, for instance in the case of a
-// timeout error. Errors can be tested against the RunError type.
-func (context *Context) Run(addressesToData map[string]any, timeout time.Duration) (matches []byte, actions []string, err error) {
-	if len(addressesToData) == 0 {
+// RunAddressData provides address data to the Context.Run method.
+type RunAddressData struct {
+	// persistent address data is scoped to the lifetime of a given Context, and subsquent calls to Context.Run with the
+	// same address name will be silently ignored.
+	persistent map[string]any
```
Hellzy, line 61: We probably want to access this from dd-trace-go when preparing the WAF call

context.go
```diff
@@ -54,12 +54,26 @@ func NewContext(handle *Handle) *Context {
...
+	// ephemeral address data is scoped to a given Context.Run call and is not persisted across calls. This is used for
+	// protocols such as gRPC client/server streaming or GraphQL, where a single request can incur multiple subrequests.
+	ephemeral map[string]any
```
Hellzy, line 64: We probably want to access this one as well

context.go
RomainMuller: There's an existing test that asserts correct behavior with `nil` addresses, and I think we should preserve the same semantics (which is, it's okay and it's a no-op).

context.go
RomainMuller: Same as https://github.com/DataDog/go-libddwaf/pull/39#discussion_r1381969290 -- I think that is not an issue?

Anilm3:
> * As you can see in the `run()` function, we keep track of the ddwaf_object parameter in the context struct because it is required for them to live all the way to the end of the context. My guess is that this is not the case for ephemeral addresses. Can you take a look at this with @Anilm3 ?

Ephemeral addresses are freed at the end of `ddwaf_run`.

RomainMuller:
> * Can you put the right permissions back on the embedded files ? (755 instead of 644)

Why'd they need to be executable? They're embedded as a byte slice in another file, and the file permissions won't be persisted when doing that anyway?

context.go
eliottness: I thought this was not coveree. Thanks for the explanation

_tools/libddwaf-updater/update.go
Julio-Guerra, line 1: I would advise using higher-level Go packages to make it closer to scripting languages and not lose the flexibility and simplicity we had. Overall I find complexity vs problem being solved very unbalanced here in the end.
Said otherwise: I'm not convinced this is a win given the higher cost of entry and overall software complexity for such a simple "copy/paste script".

Julio-Guerra: By higher-level Go packages I mean things like https://github.com/bitfield/script (never used, just a random example).

eliottness:
> > * Can you put the right permissions back on the embedded files ? (755 instead of 644)
> 
> Why'd they need to be executable? They're embedded as a byte slice in another file, and the file permissions won't be persisted when doing that anyway?

@RomainMuller This is very useful to link the library directly when debugging with a small C script. That's usually why we need to put it in 755. But more than that, it's to avoid dealing with the wierdest errors coming from the linker trying to access the library (`./a.out: permission denied` or `./a.out: no such file or directly`)

RomainMuller:
> @RomainMuller This is very useful to link the library directly when debugging with a small C script. That's usually why we need to put it in 755.

That I can understand... even though this sounds like a micro-improvement.

> But more than that, it's to avoid dealing with the wierdest errors coming from the linker trying to access the library (`./a.out: permission denied` or `./a.out: no such file or directly`)

That is only relevant when trying to directly link to those files locally, and won't change a thing with the embed (the permissions of the file we dump in a temporary location is what matters here).

_tools/libddwaf-updater/update.go
```diff
@@ -0,0 +1,284 @@
...
+		panic(err)
+	}
+	sum, err := script.File(path.Join(tmpdir, tarName)).SHA256Sum()
+	if err != nil {
+		panic(err)
+	}
+	// To match the shasum format...
+	sum = fmt.Sprintf("%s  %s\n", sum, tarName)
+	if sum != sha {
+		panic(fmt.Errorf("checksum mismatch on %s:\nExpected %s\nActual   %s", tarUrl, sha, sum))
```
eliottness, line 167:
```suggestion
		panic(fmt.Errorf("checksum mismatch on %s:\nExpected %s\nActual  %s", tarUrl, sha, sum))
```

_tools/libddwaf-updater/update.go
```diff
@@ -0,0 +1,284 @@
...
+		}
+
+		fmt.Printf("... downloaded %s\n", destPath)
+		if _, err := script.NewPipe().WithReader(arch).WriteFile(destPath); err != nil {
+			panic(err)
+		}
+		// Make the libraries executable, as this can be useful to link directly to those objects to perform troubleshooting.
+		if err := os.Chmod(destPath, 0755); err != nil {
+			panic(err)
+		}
```
eliottness, line 213: If I am right, the current code will also make the ddwaf.h file executable. Is it fine ?

encoder_test.go
eliottness, line 1: We already have a lot of encoder tests in `waf_test.go`, yours would integrate well with those. Or you can also move all of these in this file, having every tests in the same file is remnant from the previous implementation.
here is an example of what you could add to the `TestEncoder` testsuite (not sure if it compiles tho):
```go
{
    Name: "nil map",
    Data: (map[string]any)(nil)
    ExpectedWafValueLength: 0,
    ExpectedWafValueType: wafMapType,
}
```

internal/vendor/README.md
```diff
@@ -0,0 +1,22 @@
+## Embedded WAF libraries
+
+This directory contains Datadog's WAF static libraries taken from the releases
+of https://github.com/DataDog/libddwaf
+
+### Updating
+
+In order to update the embedded libraries, ensure you have a recent `node`
```
eliottness, line 8:
```suggestion
In order to update the embedded libraries, ensure you have a recent Go
```

waf_test.go
```diff
@@ -303,6 +385,84 @@ func TestMatching(t *testing.T) {
...
+	// Ephemeral address should still match, persistent shouldn't anymore
+	res, err = wafCtx.Run(runAddresses, time.Second)
+	require.NoError(t, err)
+	require.Len(t, res.Events, 1) // 1 ephemeral
+	require.Nil(t, res.Actions)
+
+	wafCtx.Close()
+	waf.Close()
+	// Using the WAF instance after it was closed leads to a nil WAF context
+	require.Nil(t, NewContext(waf))
```
eliottness, line 463: Would you be an absolute chad and add a test about what we were talking about with anil: a test about priotities between the persistents and ephemeral addresses ? (and add it to the docs of Run() on our side as a starter)

eliottness (changes requested):
image
Thanks for all of this already

_tools/libddwaf-updater/update.go
```diff
@@ -0,0 +1,280 @@
...
+	}
+
+	fmt.Println("All done! Don't forget to check in changes to include/ and internal/vendor/, check the libddwaf upgrade guide to update bindings!")
+}
+
+func createEmbedSource(tgt target) {
+	ext := "so"
+	if tgt.os == "darwin" {
+		ext = "dylib"
+	}
```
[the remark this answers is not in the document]
eliottness: Saw what you did in #40, this is very inconvinent. I will just close this

_tools/libddwaf-updater/update.go
RomainMuller: Copied from my (original) version, aligned... shows that no change is necessary:

```
Expected %s
Actual   %s
```

waf_test.go
RomainMuller: Done!

eliottness (approved):
LGTM

_tools/libddwaf-updater/update.go
```diff
@@ -0,0 +1,309 @@
...
+	//// Not ready for these just yet...
+	// {os: "windows", arch: "amd64", base: "ddwaf", ext: "dll", assetLabel: "windows-x64"},
+	// {os: "windows", arch: "386", base: "ddwaf", ext: "dll", assetLabel: "windows-win32"},
+
+	//// These are currently not supported by ebitengine/purego:
+	// {os: "linux", arch: "armv7", ext: "so", assetLabel: "armv7-linux-musl"},
+	// {os: "linux", arch: "i386", ext: "so", assetLabel: "i386-linux-musl"},
+}
+
+const vedorMarkerFile = `// Unless explicitly stated otherwise all files in this repository are licensed
```
Hellzy, line 276:
```suggestion
const vendorMarkerFile = `// Unless explicitly stated otherwise all files in this repository are licensed
```

_tools/libddwaf-updater/update.go
```diff
@@ -0,0 +1,309 @@
...
+		embedDir := path.Join(libDir, fmt.Sprintf("%s-%s", tgt.os, tgt.arch))
+		created := false
+		if _, err = os.Stat(embedDir); errors.Is(err, os.ErrNotExist) {
+			if err = os.MkdirAll(embedDir, 0755); err != nil {
+				panic(err)
+			}
+			created = true
+		}
+		if created || force {
+			if err := os.WriteFile(path.Join(embedDir, "vendor.go"), []byte(vedorMarkerFile), 0644); err != nil {
```
Hellzy, line 75:
```suggestion
			if err := os.WriteFile(path.Join(embedDir, "vendor.go"), []byte(vendorMarkerFile), 0644); err != nil {
```

Hellzy (commented):
LGTM. Typo nit

context.go
```diff
@@ -85,12 +113,12 @@ func (context *Context) Run(addressesToData map[string]any, timeout time.Duratio
 
 	// Save the Go pointer references to addressesToData that were referenced by the encoder
 	// into C ddwaf_objects. libddwaf's API requires to keep this data for the lifetime of the ddwaf_context.
-	defer context.cgoRefs.append(encoder.cgoRefs)
+	defer context.cgoRefs.append(persistentEncoder.cgoRefs)
```
Julio-Guerra, line 116: What allows us not to append `ephemeralData.cgoRefs`?
IIUC, we can avoid it because the Go reference is kept by this function call.
But if that's the case, I would try to ensure it even more explicitly to avoid potential compiler optimizations somehow removing the reference. WDYT?

Julio-Guerra: I didn't find any test on ephermeral addresses and that might be why this problem wasn't caught.

RomainMuller: There *are* definitely tests on ephemeral addresses. Not in the concurrent suite (but I also am not sure how this would uncover anything the other test hasn't).

Hellzy (approved):
I noticed you used a bunch of `interface{}` type in the tests 😛 
Not a blocker but consistency would have us switch to `any` when we can as you pointed out in other changes.

RomainMuller:
> I noticed you used a bunch of `interface{}` type in the tests 😛 Not a blocker but consistency would have us switch to `any` when we can as you pointed out in other changes.

Yeah - these were copied from surrounding style... I did not want to do a global replace as it would have further inflated the diff size...

PR #40, opened 2023-11-06
feat: add windows support

RomainMuller:


eliottness:
Sorry to break it to you but 386 in not working on purego side :sob: Here is the remnant https://github.com/ebitengine/purego/blob/main/func.go#L293. But we can still keep the CI because we want our CI to handle at least one platform not supported so we can test the unsupported code paths

.github/workflows/test.yml
```diff
@@ -97,6 +97,26 @@ jobs:
...
+      - uses: actions/setup-go@v4
+        with:
+          cache: true
+      - name: go test
+        shell: bash
+        run: |
+          # Install gotestsum
+          env GOBIN=$PWD go install gotest.tools/gotestsum@latest
+          # Run the tests with gotestsum
+          env GOARCH=${{matrix.arch}} CGO_ENABLED=${{ matrix.cgo_enabled }} ./gotestsum -- -v -count=10 -shuffle=on ./...
```
eliottness, line 118: `windows-latest` is running on amd64 right ? How do you want to run 386 tests on this ?

_tools/libddwaf-updater/update.go
```diff
@@ -207,9 +217,11 @@ func handleTarget(wg *sync.WaitGroup, version string, tgt target, embedDir strin
...
 			panic(err)
 		}
-		// Make the libraries executable, as this can be useful to link directly to those objects to perform troubleshooting.
-		if err := os.Chmod(destPath, 0755); err != nil {
-			panic(err)
+		if path.Ext(destPath) != ".h" {
+			// Make the libraries executable, as this can be useful to link directly to those objects to perform troubleshooting.
+			if err := os.Chmod(destPath, 0755); err != nil {
+				panic(err)
+			}
```
eliottness, line 224: Wrong PR :wink:

PR #42, opened 2023-11-07
support for nil and invalid value in the encoding tree

eliottness:
- [x] Replace nil errors by nil wafObject values
- [x] Replace invalid types by invalid wafObject values
- [x] Remove most errors from encoder functions
- [x] Keep map key even if the encoded value is nil or invalid
- [x] Keep struct field even if the encoded value is nil or invalid
- [x] Discard nil or invalid wafObject in arrays
- [x] Change current tests to match new implemntation
- [ ] Write Run() tests to check if the waf supported these types
- [ ] Refacto `EncoderTest` testsuite because it does not test was we would like anymore


decoder.go
```diff
@@ -127,6 +124,11 @@ func decodeDiagnosticAddresses(obj *wafObject) (res DiagnosticAddresses, err err
 }
 
 func decodeStringArray(obj *wafObject) ([]string, error) {
+	// We consider that nil is an empty array
+	if obj._type == wafNilType {
```
RomainMuller, line 128: Can you please add a
```go
func (w *wafObject) isNil() bool {
	return obj._type == wafNilType
}
```
helper?

encoder.go
```diff
@@ -30,6 +30,14 @@ type native interface {
 	int64 | uint64 | uintptr
 }
 
+func newLimitedEncoder() *encoder {
+	return &encoder{
+		containerMaxSize: wafMaxContainerSize,
+		stringMaxSize:    wafMaxStringLength,
+		objectMaxDepth:   wafMaxContainerDepth,
+	}
+}
```
RomainMuller, line 39: Out of curiosity - why does this need to be a pointer? It feels to me like this could just be returning `encoder` and it wouldn't change a thing... Same goes for `newMaxEncoder`... There is no interface involved here (that I'm aware of)...

encoder.go
```diff
@@ -38,67 +46,93 @@ func newMaxEncoder() *encoder {
 	}
 }
 
+// Encode takes a Go value and returns a wafObject pointer and an error.
+// The returned wafObject is the root of the tree of imbriqued wafObjects representing the Go value.
```
RomainMuller, line 50:
```suggestion
// The returned wafObject is the root of the tree of nested wafObjects representing the Go value.
```

encoder.go
```diff
@@ -38,67 +46,93 @@ func newMaxEncoder() *encoder {
...
+	if wo._type == wafInvalidType {
+		return nil, errUnsupportedValue
 	}
 
 	return wo, nil
 }
 
-func encodeNative[T native](val T, t wafObjectType, obj *wafObject) error {
+func encodeNative[T native](val T, t wafObjectType, obj *wafObject) {
 	obj._type = t
```
RomainMuller, line 67: Seems to me that `obj._type` could be inferred from `val.(type)` here? The current API seems susceptible to subtle mis-use.

encoder.go
```diff
@@ -38,67 +46,93 @@ func newMaxEncoder() *encoder {
...
+	return nullable && value.IsNil()
+}
+
+func (encoder *encoder) encode(value reflect.Value, obj *wafObject, depth int) {
+	value.IsValid()
 	switch kind := value.Kind(); {
 	// Terminal cases (leafs of the tree)
 	case kind == reflect.Invalid:
-		return errUnsupportedValue
+		encodeNative[uintptr](0, wafInvalidType, obj)
```
RomainMuller, line 93: Alright I guess I see why you can't infer the `obj._type` from the type argument... Somehow I dislike this particular calling pattern, but I guess... fine.

encoder.go
```diff
@@ -38,67 +46,93 @@ func newMaxEncoder() *encoder {
...
 	case kind == reflect.Map:
-		return encoder.encodeMap(value, obj, depth)
+		encoder.encodeMap(value, obj, depth)
 	case kind == reflect.Struct:
-		return encoder.encodeStruct(value, obj, depth)
+		encoder.encodeStruct(value, obj, depth)
 
 	default:
-		return errUnsupportedValue
+		encodeNative[uintptr](0, wafInvalidType, obj)
```
RomainMuller, line 126: The case where `kind == reflect.Invalid` has the exact same code as the `default` case... They could be de-duplicated.

encoder.go
```diff
@@ -127,9 +161,11 @@ func getFieldNameFromType(field reflect.StructField) (string, bool) {
...
 // - If the field has a json tag it will become the field name
 // - Private fields and also values producing an error at encoding will be skipped
-func (encoder *encoder) encodeStruct(value reflect.Value, obj *wafObject, depth int) error {
+// - Even if the element values are invalid or null we still keep them to report the field name
+func (encoder *encoder) encodeStruct(value reflect.Value, obj *wafObject, depth int) {
 	if depth < 0 {
-		return errMaxDepth
+		encodeNative[uintptr](0, wafInvalidType, obj)
+		return
 	}
```
RomainMuller, line 169: Uhhhh negative depth sounds gross to me... I'd have made it unsigned and considered `0` as "tell me there's an object, but not what's in it".

Returning `wafInvalidType` here mis-represents the fact that the value is actually valid, but we have exceeded the encoding depth and are cutting it at this stage... which I don't love (troubleshooting will be more difficult).

encoder.go
eliottness: Not mentioning that the `default` is also future-proof, the real reason we still keep the `kind == reflect.Invalid` case is because every other call on the value will potentially crash as the doc of reflect states that "each function that can be called with invalid values are explicitly stated as such, other functions do not support these values"

encoder.go
eliottness: I can't because of invalid and nil types

encoder.go
eliottness: right

encoder.go
eliottness: Fixed this here: https://github.com/DataDog/go-libddwaf/blob/e690a56d03cf815bf61978842aca8b311e47e910/encoder.go#L124

encoder.go
eliottness: refactored this

encoder_decoder_test.go
```diff
@@ -0,0 +1,716 @@
...
+			ExpectedValue map[string][]string
+		}{
+			{
+				Name:          "empty",
+				ExpectedValue: map[string][]string{},
+			},
+			{
+				Name: "one-empty-entry",
+				ExpectedValue: map[string][]string{
+					"afasdfafs": nil,
```
Hellzy, line 689: love the 🛼 ⌨️

eliottness: This one is not even from me 🤣 it's from you

https://github.com/DataDog/go-libddwaf/blob/main/decoder_test.go

waf_test.go
```diff
@@ -595,10 +595,10 @@ func TestConcurrency(t *testing.T) {
 							"user-agent": userAgents[i],
 						},
 					}
-					ephemeralData := map[string]interface{}{
+					output := map[string]interface{}{
 						"server.request.body": bodies[i],
 					}
-					res, err := wafCtx.Run(RunAddressData{Persistent: data, Ephemeral: ephemeralData}, time.Minute)
+					res, err := wafCtx.Run(RunAddressData{Persistent: data, Ephemeral: output}, time.Minute)
```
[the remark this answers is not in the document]
eliottness: The sed function took too much it, reverting it

Hellzy (approved):
Looks good on my end. Let's wait for Romain's final review as well

ctypes.go
```diff
@@ -58,6 +58,11 @@ type wafObject struct {
 	// packed (apart from breaking all tracers of course)
 }
 
+// isNil determines whether this WAF Object is nil or not.
+func (w *wafObject) isNil() bool {
+	return w._type == wafNilType
+}
```
RomainMuller, line 64: I'm torn whether this naming (which sounds more like go), and keeping the `libddwaf` semantics of `isNull` / `wafNullType`.

RomainMuller (approved):

PR #44, opened 2023-11-09
feat: improved WAF init error reporting

RomainMuller:
The `ddwaf_init` call attempts to provide more information about failure causes through the `diagnostics` object returned. This may be populated even when receiving a `NULL` handle.

This PR attempts to obtain an error from diagnostics when available in order to provide improved error messaging to the callers.


waf.go
```diff
@@ -35,14 +35,51 @@ type Diagnostics struct {
...
+		"scanners":       d.Scanners,
+	}
+
+	errs := make([]error, 0, len(fields))
+	for field, entry := range fields {
+		if entry == nil || entry.Error == "" {
+			// No entry or no error => we're all good.
+			continue
+		}
+		errs = append(errs, fmt.Errorf("in %#v: %s", field, entry.Error))
```
Julio-Guerra, line 58: It's possible to leverage an API like https://pkg.go.dev/errors#Join too for instance (our Go min version prevents us from using it)

RomainMuller: I know... Made this because I think it's a better DevX in case of issue at this stage...

handle.go
```diff
@@ -65,21 +65,41 @@ func NewHandle(rules any, keyObfuscatorRegex string, valueObfuscatorRegex string
...
+	// The WAF successfully initialized at this stage... err is the diagnostics decode error if any.
 	if err != nil {
+		wafLib.wafDestroy(cHandle)
 		return nil, fmt.Errorf("could not decode the WAF diagnostics: %w", err)
 	}
 
-	defer wafLib.wafObjectFree(diagnosticsWafObj)
+	if err = diags.TopLevelError(); err != nil {
+		defer wafLib.wafDestroy(cHandle)
+		return nil, err
```
Julio-Guerra, line 101: I'm in favor of keeping the handle in this case to allow using the WAF handle for partially applied rules:
```suggestion
		return nil, err
```

Julio-Guerra (changes requested):
Requires some adjustement to allow using a handle in case of a partially accepted rule.

waf_dl.go
```diff
@@ -99,11 +99,15 @@ func (waf *wafDl) wafGetVersion() string {
 	return gostring(cast[byte](waf.syscall(waf.getVersion)))
 }
 
-func (waf *wafDl) wafInit(ruleset *wafObject, config *wafConfig, info *wafObject) wafHandle {
+// wafInit initializes a new WAF with the provided ruleset, configuration and info objects. A
+// cgoRefPool ensures that the provided input values are not moved or garbage collected by the Go
+// runtime during the WAF call.
+func (waf *wafDl) wafInit(ruleset *wafObject, config *wafConfig, info *wafObject, refs cgoRefPool) wafHandle {
```
eliottness, line 105: If you do it here you can also do it in `wafRun` I guess

RomainMuller: Except wafRun requires some of the data to live as long as the context does, but not all of them... I actually have some (perhaps better?) ideas there, so I'll probably end up making another PR with changes to the `cgoRefPool` ergonomics...

decoder.go
```diff
@@ -112,14 +112,16 @@ func decodeDiagnosticsEntry(obj *wafObject) (*DiagnosticEntry, error) {
...
-func decodeDiagnosticAddresses(obj *wafObject) (res DiagnosticAddresses, err error) {
+func decodeDiagnosticAddresses(obj *wafObject) (res *DiagnosticAddresses, err error) {
 	if !obj.isMap() {
 		return res, errInvalidObjectType
 	}
 	if obj.value == 0 && obj.nbEntries > 0 {
 		return res, errNilObjectPtr
 	}
 
+	res = &DiagnosticAddresses{}
```
Julio-Guerra, line 123: Any chance to return `nil` when no need to have a value?
I also think it would simplify the `switch...case` block when directly returning the values from each `case` block.

handle.go
```diff
@@ -65,22 +65,37 @@ func NewHandle(rules any, keyObfuscatorRegex string, valueObfuscatorRegex string
 
 	config := newConfig(&encoder.cgoRefs, keyObfuscatorRegex, valueObfuscatorRegex)
 	diagnosticsWafObj := new(wafObject)
+	defer func() {
+		if !diagnosticsWafObj.isInvalid() {
+			wafLib.wafObjectFree(diagnosticsWafObj)
+		}
```
Julio-Guerra, line 71: Nit: create a wafObjectFree helper doing this check. I also think the WAF already handles the invalid condition.

handle.go
```diff
@@ -65,22 +65,37 @@ func NewHandle(rules any, keyObfuscatorRegex string, valueObfuscatorRegex string
...
+		diags, err = decodeDiagnostics(diagnosticsWafObj)
+	}
 
-	cHandle := wafLib.wafInit(obj, config, diagnosticsWafObj)
-	keepAlive(encoder.cgoRefs)
-	// Note that the encoded obj was copied by libddwaf, so we don't need to keep them alive
-	// for the lifetime of the handle (ddwaf API guarantee).
 	if cHandle == 0 {
+		// WAF Failed initialization, report the best possible error...
+		if err == nil && diags != nil {
```
Julio-Guerra, line 83: `err` is not clearly coming from line 78 to me when `!diagnosticsWafObj.isInvalid()` is false.
I think you can safely remove the `err == nil` part of the condition because `diags` is `nil` by default:
```suggestion
		if diags != nil {
```

handle.go
```diff
@@ -65,22 +65,37 @@ func NewHandle(rules any, keyObfuscatorRegex string, valueObfuscatorRegex string
...
+			if err != nil {
+				return nil, fmt.Errorf("could not instantiate the WAF: %w", err)
+			}
+		}
 		return nil, errors.New("could not instantiate the WAF")
 	}
 
-	diags, err := decodeDiagnostics(diagnosticsWafObj)
+	// The WAF successfully initialized at this stage... err is the diagnostics decode error if any.
 	if err != nil {
```
Julio-Guerra, line 94: Same comment here on the fact `err`'s source is conditional (on something unclear).

handle_test.go
```diff
@@ -0,0 +1,121 @@
...
+			"rules_data",
+			"rules_override",
+			"processors",
+			"scanners",
+		} {
+			t.Run(field, func(t *testing.T) {
+				// Start off with a perfectly valid input...
+				rules := makeValidRuleset()
+				// And corrupt data for one particular field...
+				rules[field] = 1337.42
```
Julio-Guerra, line 42: hack3r

handle_test.go
```diff
@@ -0,0 +1,121 @@
...
+
+package waf
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestNewHandle(t *testing.T) {
```
Julio-Guerra, line 14: If you're up to a last mile of effort, I'd suggest testing the new operator scenario too 🙏

Julio-Guerra (approved):

PR #45, opened 2023-11-13
Update go.mod file for v2

eliottness:
And remove retract directive since we changed major


Hellzy:
Maybe we should retract 2.0.0, it won't be useable

eliottness:
- [x] Retract v2.0.0
- [x] Fix error case where we would send a nil wafObject to ddwaf_run instead of a nil value

Hellzy (approved):

PR #47, opened 2023-11-15
chore: add datadog.no_waf build tag

eliottness:
- [x] Add `datadog.no_waf` build tag in all go:build directives
- [x] Add platform section as an excuse to talk about `datadog.no_waf` build tag
- [x] Add new file to explain the new reason why the waf could be disabled
- [x] Run the CI with the disabled build tag on most cases to extend coverage to the new file
- [x] Rename "Unsupported" with "Disabled" for some identifiers as it makes more sense now

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        runs-on: [ macos-13, macos-12, macos-11, ubuntu-22.04, ubuntu-20.04, windows-latest ]
+        runs-on: [ macos-13, macos-12, macos-11, ubuntu-22.04, ubuntu-20.04 ]
         go-version: [ "1.21", "1.20", "1.19" ]
         cgo_enabled: [ "0", "1" ] # test it compiles with and without cgo
         include:
@@ -30,15 +30,37 @@ jobs:
         with:
           go-version: ${{ matrix.go-version }}
           cache: true
-      - uses: autero1/action-gotestsum@v2.0.0
+      - name: go test
+        shell: bash
+        run: |
+          # Install gotestsum
+          env GOBIN=$PWD go install gotest.tools/gotestsum@latest
+          # Run the tests with gotestsum
+          env ${{ matrix.env }} CGO_ENABLED=${{ matrix.cgo_enabled }} ./gotestsum -- -v -count=10 -shuffle=on ./...
+
+  disabled:
+    strategy:
+      fail-fast: false
+      matrix:
+        runs-on: [ windows-latest, ubuntu-latest, macos-13 ]
+        go-args: [ "-tags datadog.no_waf", "-tags go1.22" ]
+        include:
+          - runs-on: windows-latest
+            go-args: ""
+    runs-on: ${{ matrix.runs-on }}
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-go@v4
         with:
-          gotestsum_version: 1.11.0
-      - name: test enabled waf
+          go-version: 'stable' # get latest stable version from https://github.com/actions/go-versions/blob/main/versions-manifest.json
+          cache: true
+      - name: go test
         shell: bash
-        run: env ${{ matrix.env }} CGO_ENABLED=${{ matrix.cgo_enabled }} gotestsum -- -v -count=10 -shuffle=on ./...
-      - name: test disabled waf
-        shell: bash
-        run: env ${{ matrix.env }} CGO_ENABLED=${{ matrix.cgo_enabled }} gotestsum -- -v -tags datadog.no_waf -shuffle=on ./...
+        run: |
+          # Install gotestsum
+          env GOBIN=$PWD go install gotest.tools/gotestsum@latest
+          # Run the tests with gotestsum
+          ./gotestsum -- -v ${{ matrix.go-tags }} -shuffle=on ./...
 
   # Same tests but on the official golang container for linux
   golang-linux-container:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -30,14 +30,15 @@ jobs:
         with:
           go-version: ${{ matrix.go-version }}
           cache: true
-
-      - name: go test
+      - uses: autero1/action-gotestsum@v2.0.0
+        with:
+          gotestsum_version: 1.11.0
+      - name: test enabled waf
         shell: bash
-        run: |
-          # Install gotestsum
-          env GOBIN=$PWD go install gotest.tools/gotestsum@latest
-          # Run the tests with gotestsum
-          env ${{ matrix.env }} CGO_ENABLED=${{ matrix.cgo_enabled }} ./gotestsum -- -v -count=10 -shuffle=on ./...
+        run: env ${{ matrix.env }} CGO_ENABLED=${{ matrix.cgo_enabled }} gotestsum -- -v -count=10 -shuffle=on ./...
+      - name: test disabled waf
+        shell: bash
+        run: env ${{ matrix.env }} CGO_ENABLED=${{ matrix.cgo_enabled }} gotestsum -- -v -tags datadog.no_waf -shuffle=on ./...
 
   # Same tests but on the official golang container for linux
   golang-linux-container:
--- a/waf.go
+++ b/waf.go
@@ -13,18 +13,6 @@ import (
 	"github.com/hashicorp/go-multierror"
 )
 
-// WafDisabledError is a wrapper error type helping to handle the error
-// case of trying to execute this package on an unsupported target environment.
-type WafDisabledError struct {
-	error
-}
-
-// Unwrap the error and return it.
-// Required by errors.Is and errors.As functions.
-func (e *WafDisabledError) Unwrap() error {
-	return e.error
-}
-
 // Diagnostics stores the information - provided by the WAF - about WAF rules initialization.
 type Diagnostics struct {
 	Version        string
@@ -140,6 +128,7 @@ var (
 	// libddwaf's dlopen error if any
 	wafErr      error
 	openWafOnce sync.Once
+	doneWafOnce bool
 )
 
 // Load loads libddwaf's dynamic library. The dynamic library is opened only
@@ -155,7 +144,14 @@ var (
 // to remove temporary files. It is safe to continue using libddwaf in such
 // case.
 func Load() (ok bool, err error) {
+	if ok, err = Health(); !ok {
+		return false, err
+	}
+
 	openWafOnce.Do(func() {
+		defer func() {
+			doneWafOnce = true
+		}()
 		wafLib, wafErr = newWafDl()
 		if wafErr != nil {
 			return
@@ -164,13 +160,6 @@ func Load() (ok bool, err error) {
 	})
 
 	return wafLib != nil, wafErr
-}
-
-// SupportsTarget returns true and a nil error when the target host environment
-// is supported by this package and can be further used.
-// Otherwise, it returns false along with an error detailing why.
-func SupportsTarget() (bool, error) {
-	return supportsTarget()
 }
 
 var wafVersion string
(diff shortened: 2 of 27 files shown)


.github/workflows/test.yml
```diff
@@ -30,14 +30,15 @@ jobs:
         with:
           go-version: ${{ matrix.go-version }}
           cache: true
-
-      - name: go test
+      - uses: autero1/action-gotestsum@v2.0.0
```
Julio-Guerra, line 33: are you sure about this source? I don't like using small custom actions (based on the name)

Julio-Guerra: Especialy just to avoid a go get - you could also cache it instead

waf_disabled_manually.go
```diff
@@ -0,0 +1,15 @@
...
+// Manually set datadog.no_waf build tag
+//go:build datadog.no_waf
+
+package waf
+
+import (
+	"fmt"
+)
+
+var disabledWafErr = &WafDisabledError{fmt.Errorf("the WAF has been manually disabled using the `datadog.no_waf` go build tag")}
```
Julio-Guerra, line 15: You got the point of this "per tag error" to offer the right log information to the user 👏

waf.go
```diff
@@ -13,15 +13,15 @@ import (
 	"github.com/hashicorp/go-multierror"
 )
 
-// UnsupportedTargetError is a wrapper error type helping to handle the error
+// WafDisabledError is a wrapper error type helping to handle the error
 // case of trying to execute this package on an unsupported target environment.
-type UnsupportedTargetError struct {
+type WafDisabledError struct {
```
Julio-Guerra, line 18: We are losing information (and breaking the public API too):
1. `UnsupportedTargetError` can be unintended and could deserve its own error-management logic
2. `WafDisabledError` is rather more likely coming from an explicit action.

I'd suggest we keep it open with two error types, that the caller can type-assert or error.Unwrap if needed
.

Julio-Guerra: `SupportsTarget()` becomes ambigous too for the disabled case, so we should work the build tags so that the unsupported target is higher prio than the disabled case

Julio-Guerra: I suggest you review dd-trace-go's and datadog-agent's use of `Health()` and `SupportsTarget()` to check we are good and we won't fall into different/new conditions.

waf.go
```diff
@@ -13,15 +13,15 @@ import (
 	"github.com/hashicorp/go-multierror"
 )
 
-// UnsupportedTargetError is a wrapper error type helping to handle the error
+// WafDisabledError is a wrapper error type helping to handle the error
 // case of trying to execute this package on an unsupported target environment.
-type UnsupportedTargetError struct {
```
Julio-Guerra, line 18: Warning: public api change ;-)

.github/workflows/test.yml
```diff
@@ -30,14 +30,15 @@ jobs:
...
         shell: bash
-        run: |
-          # Install gotestsum
-          env GOBIN=$PWD go install gotest.tools/gotestsum@latest
-          # Run the tests with gotestsum
-          env ${{ matrix.env }} CGO_ENABLED=${{ matrix.cgo_enabled }} ./gotestsum -- -v -count=10 -shuffle=on ./...
+        run: env ${{ matrix.env }} CGO_ENABLED=${{ matrix.cgo_enabled }} gotestsum -- -v -count=10 -shuffle=on ./...
+      - name: test disabled waf
+        shell: bash
+        run: env ${{ matrix.env }} CGO_ENABLED=${{ matrix.cgo_enabled }} gotestsum -- -v -tags datadog.no_waf -shuffle=on ./...
```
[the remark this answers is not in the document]
eliottness: I don't think this is necessary because this would be very tedious and we are htting our parralel job limit for the repo with only one PR already. But this is indeed not ideal, I will create another job running only what's necessary to test this

waf.go
eliottness: Reverted the changes

waf.go
eliottness: I managed to change thing into something I am comfortable with

.github/workflows/test.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        runs-on: [ macos-13, macos-12, macos-11, ubuntu-22.04, ubuntu-20.04, windows-latest ]
```
Julio-Guerra, line 17: why? this is our unsupported target right now :-)

.github/workflows/test.yml
```diff
@@ -39,6 +38,30 @@ jobs:
           # Run the tests with gotestsum
           env ${{ matrix.env }} CGO_ENABLED=${{ matrix.cgo_enabled }} ./gotestsum -- -v -count=10 -shuffle=on ./...
 
+  disabled:
+    strategy:
+      fail-fast: false
+      matrix:
+        runs-on: [ windows-latest, ubuntu-latest, macos-13 ]
```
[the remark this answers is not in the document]
eliottness: I would reply that we don't add a new platform every now and then. And this still represents 10% of the work of moving all the build tags around. We can add a guide for this in the readme if you want

waf.go
Julio-Guerra: My bad I saw it in the new file now.

waf_support.go
```diff
@@ -0,0 +1,73 @@
+package waf
+
+import (
+	"fmt"
+
+	"github.com/hashicorp/go-multierror"
+)
+
+var wafSupportErrors []error
+var wafManuallyDisabledErr error
```
Julio-Guerra, line 10: Worth a comment explaining how do they must be used

waf_support.go
```diff
@@ -0,0 +1,73 @@
...
+
+func (e UnsupportedGoVersionError) Error() string {
+	return fmt.Sprintf("unsupported Go version: %s", e.version)
+}
+
+// ManuallyDisabledError is a wrapper error type helping to handle the error
+// case of trying to execute this package when the WAF has been manually disabled with
+// the `datadog.no_waf` go build tag.
+type ManuallyDisabledError struct {
+}
```
Julio-Guerra, line 37:
```suggestion
type ManuallyDisabledError struct {}
```

waf_support.go
```diff
@@ -0,0 +1,73 @@
...
+
+// Health returns true if the waf is usable, false otherwise. At the same time it can return an error
+// if the waf is not usable, but the error is not blocking if true is returned, otherwise it is.
+// The following conditions are checked:
+// - The Waf library has been loaded successfully (you need to call `Load()` first for this case to be taken into account)
+// - The Waf library has not been manually disabled with the `datadog.no_waf` go build tag
+// - The Waf library is not in an unsupported OS/Arch
+// - The Waf library is not in an unsupported Go version
+func Health() (bool, error) {
+
```
Julio-Guerra, line 58:
```suggestion
```

waf_support.go
```diff
@@ -0,0 +1,73 @@
...
+	if len(wafSupportErrors) > 0 {
+		err = multierror.Append(err, wafSupportErrors...)
+	}
+
+	if wafManuallyDisabledErr == nil {
+		err = multierror.Append(err, wafManuallyDisabledErr)
+	}
+
+	return (!doneWafOnce || wafLib != nil) && len(wafSupportErrors) == 0 && wafManuallyDisabledErr == nil, err.ErrorOrNil()
```
Julio-Guerra, line 72: There's a race on `doneWafOnce`. The one in wafOnce is protected but not this one, so the write and read access can happen at the same time.

waf_unsupported_test.go
```diff
@@ -20,12 +19,16 @@ func TestLoad(t *testing.T) {
...
 func TestSupportsTarget(t *testing.T) {
 	supported, err := waf.SupportsTarget()
 	require.False(t, supported)
 	require.Error(t, err)
 }
+
+func TestHealth(t *testing.T) {
+	ok, err := waf.Health()
+	require.False(t, ok)
+	require.Error(t, err)
```
Julio-Guerra, line 33: The last [big] mile of effort would be to test every single error type ;-p

waf_unsupported_target.go
```diff
@@ -4,14 +4,15 @@
 // Copyright 2016-present Datadog, Inc.
 
 // Unsupported target OS or architecture on a supported Go version
-//            Unsupported OS        Unsupported Arch      Good Go Version
-//go:build ((!linux && !darwin) || (!amd64 && !arm64)) && !go1.22
+//            Unsupported OS        Unsupported Arch   Good Go Version Not manually disabled
+//go:build ((!linux && !darwin) || (!amd64 && !arm64)) && !go1.22 && !datadog.no_waf
```
Julio-Guerra, line 8: Don't you need to relax this build condition to cumulate multiple `wafSupportErrors` cases?
To only target unsupported targets:
 ```suggestion
//go:build ((!linux && !darwin) || (!amd64 && !arm64))
```

**Review by Julio-Guerra — CHANGES_REQUESTED** [review]
I think there's a mistake in the build tags to append all the unsupported target errors cases as expected.

File: waf_unsupported_target.go (hunk 13 above)
Comment (Julio-Guerra) [reply waf_unsupported_target.go:8]: Same comment with the other files appending to this var.

Comment (Julio-Guerra) [reply waf_unsupported_target.go:8]: Same comment with the other files appending to this var.

File: waf_support.go (hunk 11 above)
Comment (eliottness) [reply waf_support.go:72]: But if it's the case here it's also the case with `wafLib` if someone calls `Load()` and  `NewHandle` at the same time. When I wrote this I tought "since it's an assignment of a native type it's therefore not subject to race conditions". I guess it's possible at some point that this value will still be false and dlopen will just have finished running but this does not seem to be concerning.

Comment (eliottness) [reply waf_support.go:72]: We can also make all variables atomic and it will be fine if you want

File: waf_unsupported_target.go (hunk 13 above)
Comment (eliottness) [reply waf_unsupported_target.go:8]: right

File: waf_unsupported_test.go (hunk 12 above)
Comment (eliottness) [reply waf_unsupported_test.go:33]: We would need to divise test files with different build tags.... meh.... (till we refactored go-libddwaf with more packages)

File: waf_support.go (hunk 11 above)
Comment (Julio-Guerra) [reply waf_support.go:72]: The API is indeed making it impossible by design to concurrently accessed.

Comment (Julio-Guerra) [reply waf_support.go:72]: isn't `wafLib != nil` enough? I'm not clear why `doneWafOnce` is needed.

File: waf_dl_unsupported.go [hunk 14]
```diff
@@ -11,7 +11,7 @@ package waf
 type wafDl struct{}
 
 func newWafDl() (dl *wafDl, err error) {
-	return nil, unsupportedTargetErr
+	return nil, nil
```
Julio-Guerra, line 14:
```suggestion
	_, err := Health()
	return nil, err
```

waf_unsupported_go.go
```diff
@@ -3,15 +3,15 @@
 // This product includes software developed at Datadog (https://www.datadoghq.com/).
 // Copyright 2016-present Datadog, Inc.
 
-// Supported OS/Arch but unsupported Go version
-//           Supported OS        Supported Arch     Bad Go Version
-//go:build (linux || darwin || windows) && (amd64 || arm64) && go1.22
+// Unsupported Go version
```
Julio-Guerra, line 6:
```suggestion
// Unsupported Go versions (>=)
```

waf_unsupported_test.go
Julio-Guerra: lol yes that's why it's a hard one - but definitely worth it given this complex (but useful) error generation that you have built.

worth a TODO otherwise:
```suggestion
	// TODO: finer-grain error checking of all the different error possibilities (unsupported go version, unsupported target, disabled, etc.)
	require.Error(t, err)
```

waf_support.go
eliottness:
```suggestion
// Errors used to report data using the Health function
// Store all the errors related to why go-liddwaf is unavailable for the current target at runtime.
var wafSupportErrors []error
// Not nil if the build tag `datadog.no_waf` is set
var wafManuallyDisabledErr error
```

Julio-Guerra (approved):

PR #48, opened 2023-11-16
rename _noop_free -> _noop_free_v2

eliottness:
Go ASM symbols are not bound by the major versioning ABI breakage go imposes, therefore we must change the name of the symbol manually for the v1 and the v2 version to coexist in the same binary.

Example error:
```
link: duplicated definition of symbol _noop_free, from github.com/DataDog/go-libddwaf/internal/noopfree and github.com/DataDog/go-libddwaf/v2/internal/noopfree
```

### DIFF
--- a/internal/noopfree/noopfree.go
+++ b/internal/noopfree/noopfree.go
@@ -10,6 +10,6 @@ package noopfree
 
 import "unsafe"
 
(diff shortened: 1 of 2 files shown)


RomainMuller (approved):
Please update the PR description to describe why you're doing this change.

PR #50, opened 2023-11-23
refactor embed storage and packaging for bazel 

eliottness:
Relates to https://github.com/bazelbuild/bazel-gazelle/issues/1316

The solution is to keep the embedded files within the same package as their embed directive.


internal/lib/lib.go
```diff
@@ -12,6 +12,13 @@ import (
 	"os"
 
 	_ "embed"
+
+	// Do not remove the following imports which allow supporting package
+	// vendoring by properly copying all underlying libraries
+	_ "github.com/DataDog/go-libddwaf/v2/internal/lib/darwin-amd64"
+	_ "github.com/DataDog/go-libddwaf/v2/internal/lib/darwin-arm64"
+	_ "github.com/DataDog/go-libddwaf/v2/internal/lib/linux-amd64"
+	_ "github.com/DataDog/go-libddwaf/v2/internal/lib/linux-arm64"
```
RomainMuller, line 21: These files are gated by a specific build tag... do these still work despite this?

Julio-Guerra: I'm especially worried by the fact all of them will be embedded if someone imports this package in go-libddwaf.
Are we avoiding it right now?

eliottness: Indeed this is the case, I guess we need to remove it 🤷

Julio-Guerra: oh you just copy/pasted old vendoring code? because it wasn't using go:embed, that's why we didn't have the issue.
Vendoring just needs a .go file to be present in the package tree, that's why the vendor.go files were enough along with the .a files.

RomainMuller (changes requested):
You need to update the https://github.com/DataDog/go-libddwaf/tree/main/_tools/libddwaf-updater also, so it generates the correct/expected files...

Probably worth adding a comment stating those files are supposed to be tool-managed/tool-updated...

Julio-Guerra (approved):

Hellzy (approved):

RomainMuller (approved):

PR #52, opened 2023-12-04
fix: adjust mutex handling around ddwaf_context

RomainMuller:
- Removes a leftover mutex on `Handle` that no longer served any purpose
- Acquiring the mutex on `Context` prior to calling `ddwaf_context_destroy` to avoid concurrent calls to `ddwaf_run`, as these are not thread-safe.

APPSEC-12585

### DIFF
--- a/context.go
+++ b/context.go
@@ -41,13 +41,13 @@ type Context struct {
 // or the WAF context couldn't be created.
 func NewContext(handle *Handle) *Context {
 	// Handle has been released
-	if handle.addRefCounter(1) == 0 {
+	if !handle.retain() {
 		return nil
 	}
 
 	cContext := wafLib.wafContextInit(handle.cHandle)
 	if cContext == 0 {
-		handle.addRefCounter(-1)
+		handle.Close() // We couldn't get a context, so we no longer have an implicit reference to the Handle in it...
 		return nil
 	}
 
@@ -120,12 +120,9 @@ func (context *Context) Run(addressData RunAddressData, timeout time.Duration) (
 	return
 }
 
+// run executes the ddwaf_run call with the provided data on this context. The caller is responsible for locking the
+// context appropriately around this call.
 func (context *Context) run(persistentData, ephemeralData *wafObject, timeout time.Duration, cgoRefs *cgoRefPool) (Result, error) {
-	// RLock the handle to safely get read access to the WAF handle and prevent concurrent changes of it
-	// such as a rules-data update.
-	context.handle.mutex.RLock()
-	defer context.handle.mutex.RUnlock()
-
 	result := new(wafResult)
 	defer wafLib.wafResultFree(result)
 
@@ -172,7 +169,12 @@ func unwrapWafResult(ret wafReturnCode, result *wafResult) (res Result, err erro
 
 // Close calls handle.closeContext which calls ddwaf_context_destroy and maybe also close the handle if it in termination state.
 func (context *Context) Close() {
-	defer context.handle.closeContext(context)
+	context.mutex.Lock()
+	defer context.mutex.Unlock()
+
+	wafLib.wafContextDestroy(context.cContext)
+	context.handle.Close() // Reduce the reference counter of the Handle.
+
 	// Keep the Go pointer references until the end of the context
 	keepAlive(context.cgoRefs)
 	// The context is no longer used so we can try releasing the Go pointer references asap by nulling them
--- a/context.go
+++ b/context.go
@@ -163,14 +163,17 @@ func unwrapWafResult(ret wafReturnCode, result *wafResult) (res Result, err erro
 	return res, err
 }
 
-// Close calls handle.closeContext which calls ddwaf_context_destroy and maybe also close the handle if it in termination state.
+// Close the underlying `ddwaf_context` and releases the associated internal
+// data. Also decreases the reference count of the `ddwaf_hadnle` which created
+// this context, possibly releasing it completely (if this was the last context
+// created from this handle & it was released by its creator).
 func (context *Context) Close() {
 	context.mutex.Lock()
 	defer context.mutex.Unlock()
 
 	wafLib.wafContextDestroy(context.cContext)
-	keepAlive(context.cgoRefs)   // Keep the Go pointer references until the end of the context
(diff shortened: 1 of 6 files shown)


eliottness (approved):

Hellzy (approved):

context.go
```diff
@@ -172,11 +165,15 @@ func unwrapWafResult(ret wafReturnCode, result *wafResult) (res Result, err erro
...
-	// Keep the Go pointer references until the end of the context
-	keepAlive(context.cgoRefs)
-	// The context is no longer used so we can try releasing the Go pointer references asap by nulling them
-	context.cgoRefs = cgoRefPool{}
+	context.mutex.Lock()
+	defer context.mutex.Unlock()
+
+	wafLib.wafContextDestroy(context.cContext)
+	keepAlive(context.cgoRefs)   // Keep the Go pointer references until the end of the context
+	defer context.handle.Close() // Reduce the reference counter of the Handle.
```
Julio-Guerra, line 173:
```suggestion
	defer context.handle.release() // Reduce the reference counter of the Handle.
```

Julio-Guerra (changes requested):

context.go
```diff
@@ -16,22 +16,18 @@ import (
...
-	cContext wafContext
-	// cgoRefs is used to retain go references to WafObjects until the context is destroyed.
-	// As per libddwaf documentation, WAF Objects must be alive during all the context lifetime
-	cgoRefs cgoRefPool
-	// Mutex protecting the use of cContext which is not thread-safe and cgoRefs.
-	mutex sync.Mutex
+	handle *Handle // Instance of the WAF
+
+	cgoRefs  cgoRefPool // Used to retain go data referenced by WAF Objects the context holds
+	cContext wafContext // The C ddwaf_context pointer
```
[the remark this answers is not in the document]
RomainMuller: The field ordering here is determined by `betteralign` to minimize the pointer surface for the Go garbage collector.

handle.go
```diff
@@ -142,36 +136,49 @@ func (handle *Handle) Update(newRules any) (*Handle, error) {
...
-		if swapped := handle.refCounter.CompareAndSwap(current, current+x); swapped {
-			return current + x
+
+		next := current + x
+		if swapped := handle.refCounter.CompareAndSwap(current, next); swapped {
+			if next < 0 {
+				// Somehow, the counter got negative, which should never happen if
+				// `addRefCount` is called only with +1 or -1. In any case, we return 0
+				// to signifiy we reached 0 for the first time right now.
+				return 0
```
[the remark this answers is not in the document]
RomainMuller: I'd panic if this didn't mean possibly crashing a customer application because of a WAF bug...?

waf_test.go
```diff
@@ -817,6 +817,52 @@ func TestConcurrency(t *testing.T) {
...
+			},
+		}
+
+		for n := 0; n < nbUsers; n++ {
+			n := n
+			go func() {
+				startBarrier.Wait()
+				defer stopBarrier.Done()
+
+				time.Sleep(time.Microsecond)
```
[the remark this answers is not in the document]
RomainMuller: This gives us CPU-scheduler-supported randomness of execution order :)

context.go
```diff
@@ -120,12 +116,9 @@ func (context *Context) Run(addressData RunAddressData, timeout time.Duration) (
...
 }
 
+// run executes the ddwaf_run call with the provided data on this context. The caller is responsible for locking the
+// context appropriately around this call.
 func (context *Context) run(persistentData, ephemeralData *wafObject, timeout time.Duration, cgoRefs *cgoRefPool) (Result, error) {
-	// RLock the handle to safely get read access to the WAF handle and prevent concurrent changes of it
-	// such as a rules-data update.
-	context.handle.mutex.RLock()
-	defer context.handle.mutex.RUnlock()
-
```
Julio-Guerra, line 128: Indeed no longer needed because we are by construction only reading the handle once instantiated.

context.go
Julio-Guerra, line 173:
```suggestion
	defer context.handle.release() // Reduce the reference counter of the Handle.
```

context.go
```diff
@@ -172,11 +165,15 @@ func unwrapWafResult(ret wafReturnCode, result *wafResult) (res Result, err erro
 
 // Close calls handle.closeContext which calls ddwaf_context_destroy and maybe also close the handle if it in termination state.
```
Julio-Guerra, line 166:
```suggestion
// Close the ddwaf_context and associated internal data. Also decreases the number of
// references to the ddwaf_handle of the context, which possibly releases it if this is the
// last context referencing it.
```

context.go
```diff
@@ -172,11 +165,15 @@ func unwrapWafResult(ret wafReturnCode, result *wafResult) (res Result, err erro
...
 func (context *Context) Close() {
-	defer context.handle.closeContext(context)
-	// Keep the Go pointer references until the end of the context
-	keepAlive(context.cgoRefs)
-	// The context is no longer used so we can try releasing the Go pointer references asap by nulling them
-	context.cgoRefs = cgoRefPool{}
+	context.mutex.Lock()
+	defer context.mutex.Unlock()
```
Julio-Guerra, line 169: So this is where the mutex lock was forgotten indeed. Very good catch.

Note: having concurrent calls to wafCtx.Run and wafCtx.Close doesn't happen in the current way we instrument. But it's definitely worth making it correct.

handle.go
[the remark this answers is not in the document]
Julio-Guerra: It's something I always struggle to decide. But I like to raise something in case of unexpected behaviours, such as a telemetry metric, an error log, a panic, etc.

One of our tasks in the backlog is to add panic recovering in all our go-libddwaf API to be able to gracefully return an error in case of a panic (and gracefully try to handle it).

My conclusion is: if you agree this case is an unexpected behaviour, let's at least add a TODO saying to raise some kind of future signal (error, panic, whatever).

Julio-Guerra (changes requested):
Full pass done.
Please let me know when you need a final review :-)

Julio-Guerra (approved):

PR #54, opened 2023-12-13
Improve test matrix comprehensiveness

RomainMuller:
Run tests with a comprehensive cartesian product of all configurations:
- Operating systems & architectures[^1]
- CGO_ENABLED or not
- Go runtimes: `1.19`, `1.20`, `1.21`
- Go tags: _none_, `datadog.no_waf`, `go1.22`[^2], `datadog.no_waf,go1.22`

This uncovered compatibility gaps in the matrix, which this PR fixes as well:
- unsupported platform or architecture + `datadog.no_waf`
- unsupported go runtime + unsupported platform

[^1]: Multiple architectures are only tested for linux platforms due to technical limitations
[^2]: `go1.22` is a stand-in for future go releases for which `purego` support needs checking

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -100,10 +100,10 @@ jobs:
           # Prune inexistant build images (debian buster is on LTS but won't get new go version images)
           - go-version: '1.21'
             image: golang:{0}-buster
-          # The amazonlinux:2 variant is only relevant for the default go version yum ships
+          # The amazonlinux:2 variant is only relevant for the default go version yum ships (currently 1.20)
           - go-version: '1.19'
             image: amazonlinux:2
-          - go-version: '1.20'
+          - go-version: '1.21'
             image: amazonlinux:2
     name: linux/${{ matrix.arch }} ${{ format(matrix.image, matrix.go-version) }} cgo=${{ matrix.cgo-enabled }} tags=${{ matrix.go-tags }}
     runs-on: ubuntu-latest
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -111,7 +111,7 @@ jobs:
         run: |-
           docker run --name gha-${{ github.run_id }} --rm -di                   \
             --platform="linux/${{ matrix.arch }}"                               \
-            -v "${HOME}/go/pkg/mod:/root/go/pkg/mod"                            \
+            -v "${HOME}/go/pkg/mod:/go/pkg/mod"                                 \
             -v "$PWD:$PWD"                                                      \
             -w "$PWD"                                                           \
             -eCGO_ENABLED="${{ matrix.cgo-enabled }}"                           \
@@ -121,14 +121,18 @@ jobs:
             -eLD_DEBUG=all                                                      \
             "golang:${{ matrix.go-version }}-${{ matrix.base }}"
       - name: Install alpine requirements
-        if: matrix.base == 'alpine'
+        if: matrix.base == 'alpine' && matrix.cgo-enabled == '1'
         run: |-
           docker exec -i gha-${{ github.run_id }}                               \
             apk add gcc musl-dev libc6-compat
+      - name: Install gotestsum
+        run: |-
+          docker exec -i gha-${{ github.run_id }}                               \
+            go install gotest.tools/gotestsum@latest
       - name: go test
         run: |-
           docker exec -i gha-${{ github.run_id }}                               \
-            go test -v -count=10 -shuffle=on -tags='${{ matrix.go-tags }}' ./...
+            gotestsum -- -v -count=10 -shuffle=on -tags='${{ matrix.go-tags }}' ./...
       - name: Stop container
         if: always() && steps.container.outcome == 'success'
         run: |-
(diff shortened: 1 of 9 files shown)


.github/workflows/test.yml
```diff
@@ -7,127 +7,143 @@ on:
...
-        distribution: [ bookworm, bullseye, buster, alpine ]
-        cgo_enabled: [ "0", "1" ] # test it compiles with and without cgo
+        cgo-enabled: [ "0", "1" ] # test it compiles with and without cgo
+        go-tags:
+          - ''                      # Default behavior
+          - 'datadog.no_waf'        # Explicitly disabled WAF
+          - 'go1.22'                # Too recent go version (purego compatibility uncertain)
+          - 'datadog.no_waf,go1.22' # Explicitly disabled & too recent go version (purego compatibility uncertain)
+        include:
+          # gocheck2 is configrued differently in go1.21 than in previous versions
```
eliottness, line 76:
```suggestion
          # gocheck2 is configured differently in go1.21 than in previous versions
```

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/DataDog/go-libddwaf/v2
 
-go 1.18
+go 1.19
```
eliottness, line 3: Has nothing to do in this PR, same for the filename change but we can let it slide I guess

waf_manually_disabled.go
```diff
@@ -4,7 +4,7 @@
 // Copyright 2016-present Datadog, Inc.
 
 // Manually set datadog.no_waf build tag
-//go:build datadog.no_waf
+//go:build datadog.no_waf && (linux || darwin) && (amd64 || arm64) && !go1.22
```
eliottness, line 7: This (as well as the other build tag changes) remove the feature of been able to report multiple compatibility issues at the same time. How about fixing the issues in the tests rather than changing the build tags on the reporters?

waf_unsupported_target.go
```diff
@@ -5,7 +5,7 @@
 
 // Unsupported target OS or architecture
 //            Unsupported OS        Unsupported Arch
-//go:build (!linux && !darwin) || (!amd64 && !arm64)
+//go:build ((!linux && !darwin) || (!amd64 && !arm64)) && !go1.22
```
eliottness, line 8: same as before

waf_unsupported_target_test.go
```diff
@@ -3,15 +3,16 @@
 // This product includes software developed at Datadog (https://www.datadoghq.com/).
 // Copyright 2016-present Datadog, Inc.
 
-//go:build (!linux && !darwin) || (!amd64 && !arm64)
+//go:build ((!linux && !darwin) || (!amd64 && !arm64)) && !go1.22
```
eliottness, line 6: same as before but with the extra spicy sauce about us needing to uniformize formatter config

eliottness (changes requested):
I still don't get why you needed to change the job names, they were explicit enough I think

waf_manually_disabled.go
RomainMuller: > remove the feature of been able to report multiple compatibility issues at the same time

That feature never worked because all the variations end up defining some identical symbols, so you can't have all of them in at the same time. So sure, we could instead change code so that it's possible to have those cases enabled at the same time, by de-dupliciating the symbols in question... but then that would be ADDING a feature, and not fixing what was before :)

RomainMuller: I stand corrected - the code itself worked fine, only the tests were breaking due to conflicting symbols. That is an easy fix actually.

.github/workflows/test.yml
```diff
@@ -7,127 +7,163 @@ on:
...
+      - name: Install gotestsum
+        run: |-
+          docker exec -i gha-${{ github.run_id }}                               \
+            go install gotest.tools/gotestsum@latest
+      - name: go test
+        run: |-
+          docker exec -i gha-${{ github.run_id }}                               \
+            go run gotest.tools/gotestsum@latest --                             \
+              -v -count=10 -shuffle=on -tags='${{ matrix.go-tags }}'             \
+              ./...
```
Julio-Guerra, line 154: I usually prefer a multi-stage Dockerfile where the final container image only contains the produced binary and its dependencies, without all of the dev tools. This allows us to find setup limitations/issues.

OTOH, it would be valid to say this test's role is just to execute the tests on the given containers, regardless of the setup/onboarding use-cases.

Julio-Guerra: But what I also think is that setup cases should be tested in the appsec test app https://github.com/DataDog/appsec-go-test-app/blob/main/Dockerfile at the very least

RomainMuller: I'd need to figure out how to pre-build test binaries and correctly execute them then... but I'm not against the idea. We'd then also want bare-metal tests to have separated build & run jobs (where run only has the test binary available).

RomainMuller:
>I still don't get why you needed to change the job names, they were explicit enough I think

I largely re-made this from scratch and initially wanted to combine all flavors in a single job (but that is terribly inconvenient to mix docker & bare metal scenarios)

Julio-Guerra (approved):

eliottness (approved):
<h1>:+1:</h1>

PR #56, opened 2023-12-15
feat: DD_APPSEC_WAF_LOG_LEVEL can be used to enable libddwaf logging

RomainMuller:
Setting `DD_APPSEC_WAF_LOG_LEVEL` to a valid log level will result in registering a log callback with `libddwaf` handles that prints log messages it emits using the `log` standard library package.

Valid log levels are (case-insensitive) `TRACE`, `DEBUG`, `INFO`, `WARN`, `ERROR`, and `OFF` (which registers the callback, but does not result in anything being printed).

The `DD_WAF_LOG_LEVEL` setting is inoperant if CGO is disabled (`CGO_ENABLED=0`) and the current platform (OS + Architecture + Go Runtime) does not have `purego.NewCallback` support (e.g: on all `linux` OSes).

When `DD_APPSEC_WAF_LOG_LEVEL` is defined (and not blank), `DD_APPSEC_WAF_LOG_FILTER` can be set to a regular expression to filter logging messages to only those matched by the regular expression.


eliottness:
1) if you want tu use callbacks you will need to upgrade purego to 0.5.1

2) I am against the use of purego callbacks as they are too much unstable

3) Please make sure that the datadog.no_waf build tag will disable all occurrences of cgo and purego.

4) the refacto of the github workflows seems widely unnecessary.

RomainMuller:
>the refacto of the github workflows seems widely unnecessary.

Oh - ouch. This was indeed intended to be a separate PR...

RomainMuller:
>I am against the use of purego callbacks as they are too much unstable

The test matrix demonstrates this works as intended... On top of this, the Go runtime is unlikely to mess it up by moving the memory being used around while in C-world as this is operating solely on strings owned by C which the Go runtime doesn't tamper with.

eliottness:
> The test matrix demonstrates this works as intended... On top of this, the Go runtime is unlikely to mess it up by moving the memory being used around while in C-world as this is operating solely on strings owned by C which the Go runtime doesn't tamper with.
Unfortunately the test matrix can't find everything. The whole feature on purego side has a lot of assembly code and we are not covered from cases like [this one](https://github.com/ebitengine/purego/issues/106) from some months ago where the issue was the version of the glibc used. 
But I can reformulate: I am against introducing this feature of purego unless we perform extensive QA beforehand to make sure this new assembly code is safe. Moreover this is unnecessary as the main goal is to log output in datadog-agent which uses cgo. So there is no direct benefit from introducing this here.

RomainMuller:
> the main goal is to log output in datadog-agent

Also in `dd-trace-go`...

eliottness:
Since it's is required to set the env var them I am okay with this

eliottness:
I hate to have dangling helper functions everywhere, especially unsafe ones like gostring. Let's create a package `internal/unsafe` when we have some time

eliottness (approved):

PR #57, opened 2023-12-18
 refactor!: opt-in with the appsec build tag when cgo is disabled

eliottness:


waf_cgo_disabled.go
```diff
@@ -0,0 +1,13 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-present Datadog, Inc.
+
+// Unsupported Go versions (>=)
```
Julio-Guerra, line 6:
```suggestion
```

Julio-Guerra, line 6:
```suggestion
// The Go build tag "appsec" was introduced to avoid having CGO_ENABLED=0 breaking changes
// due to purego's dynamic link against libdl.so, which is not expected when CGO is disabled. 
```

waf_support.go
```diff
@@ -38,6 +38,12 @@ func (e UnsupportedGoVersionError) Error() string {
 	return fmt.Sprintf("unsupported Go version: %s", runtime.Version())
 }
 
+type CgoDisabledError struct{}
+
+func (e CgoDisabledError) Error() string {
+	return "the WAF is disabled by default when CGO is disabled, to enable it you can still use the build tag `appsec`"
```
Julio-Guerra, line 44:
```suggestion
	return "go-libddwaf is disabled when cgo is disabled unless you compile with the go build tag `appsec`. It will require libdl.so.2 at run time"
```

eliottness:
```suggestion
	return "go-libddwaf is disabled when cgo is disabled unless you compile with the go build tag `appsec`. It will require libdl.so.2. libpthread.so.0, libc.so.6 and libm.so.6 shared libraries at run time on linux"
```

Julio-Guerra (approved):

PR #59, opened 2024-01-04
add go1.22-rc docker images to the test matrix

eliottness:
Prepare for go 1.22 release

### DIFF
--- a/.github/workflows/_test_containerized.yml
+++ b/.github/workflows/_test_containerized.yml
@@ -16,7 +16,7 @@ jobs:
           # RPM-based image
           - amazonlinux:2 # pretty popular on AWS workloads
         arch: [ amd64, arm64 ]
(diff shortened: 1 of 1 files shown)


.github/workflows/_test_containerized.yml
```diff
@@ -16,7 +16,7 @@ jobs:
           # RPM-based image
           - amazonlinux:2 # pretty popular on AWS workloads
         arch: [ amd64, arm64 ]
-        go-version: [ "1.21", "1.20", "1.19" ]
+        go-version: [ "1.22-rc", "1.21", "1.20", "1.19" ]
```
RomainMuller, line 19: All other places in this file which refer to 1.21 probably also need a 1.22 version (`include`/`exclude`).

RomainMuller (changes requested):
We can also test in the non-containerized form...

PR #68, opened 2024-02-08
fix: ignore byte array fields and add ddwaf:ignore struct tag

eliottness:


context.go
```diff
@@ -70,6 +70,7 @@ func (d RunAddressData) isEmpty() bool {
 // matches as a JSON string (usually opaquely used) along with the corresponding actions in any. In case of an error,
 // matches and actions can still be returned, for instance in the case of a timeout error. Errors can be tested against
 // the RunError type.
+// Struct fields having the tag `ddwaf:ignore` will not be encoded and sent to the WAF
```
RomainMuller, line 73: That should be `ddwaf:"ignore"`, I think. That feature also appears untested... SMH.

eliottness: Oops, forgot to include them, here they are

RomainMuller: Still,

```suggestion
// Struct fields having the tag `ddwaf:"ignore"` will not be encoded and sent to the WAF
```

RomainMuller (approved):

RomainMuller (approved):

RomainMuller (approved):

PR #69, opened 2024-02-08
refactor: better packaging

eliottness:
Create new packages to un-bloat the main package:
* `internal/bindings`: low-level bindings, mirroring of ddwaf.h types in Go in ctypes.go
* `internal/unsafe`: helpers to centralize and control the use of the unsafe package
* `internal/support`: used to determine (with build tags) if the current platform and compilation parameters are supported by the WAF
* `errors`: Put in a separate package the custom error types that can be returned by the WAF

Leaves only the core code in the main package


eliottness:
Back to draft, it's to taxing to work we the current CI

RomainMuller (approved):

PR #70, opened 2024-02-12
feat(encoder): coerce aliases of byte slices as byte slices

eliottness:
One of the sources of false positives are partial parsing using `json.RawMessage` for json parsing for example. Ignoring these kind of values during encoding makes the WAF more reliable and faster in case different types of byte slice aliases are present


Julio-Guerra (approved):
LGTM

PR #71, opened 2024-02-12
ci: tranfer part of the matrix to a shell script

eliottness:
- [x] Run part of the previous job matrix as a script shell.
- [x] Switch go 1.22rc2 in favor of 1.22
- [x] Overall bump all timeout values for less flakyness 
- [x] Added missing build tags on handle_test.go
- [x] remove gotestsum 

Pros:
* Decreased CI time: p50 ~ 30min -> 3min | p95 ~ 2h -> 10m
* More understandable
* Does not hit the github job limit if we need to add another OS or Arch
* Less arm64 jobs = overall less flakyness

Cons:
* shell
* Alpine packages and Windows makes the shell a little tricky
* We are still dependent on arm64 job length

### DIFF
--- a/.github/workflows/_test_bare_metal.yml
+++ b/.github/workflows/_test_bare_metal.yml
@@ -8,7 +8,7 @@ jobs:
       fail-fast: false
       matrix:
         runs-on: [ macos-14, macos-13, macos-12, macos-11, ubuntu-22.04, ubuntu-20.04, ubuntu-latest-16-cores, windows-latest ]
-        go-version: [ "1.22.0", "1.21", "1.20", "1.19" ]
+        go-version: [ "1.22", "1.21", "1.20", "1.19" ]
         include:
           # Test with DD_APPSEC_WAF_LOG_LEVEL (only latest go, without any particular tag)
           - go-version: '1.22'
--- a/.github/workflows/_test_bare_metal.yml
+++ b/.github/workflows/_test_bare_metal.yml
@@ -7,10 +7,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        runs-on: [ macos-14, macos-13, macos-12, macos-11, ubuntu-22.04, ubuntu-20.04, ubuntu-latest-16-cores, windows-latest ]
+        runs-on: [ macos-14, macos-13, macos-12, macos-11, ubuntu-22.04, ubuntu-20.04, windows-latest ]
         go-version: [ "1.22", "1.21", "1.20", "1.19" ]
         include:
-          # Test with DD_APPSEC_WAF_LOG_LEVEL (only latest go, without any particular tag)
+          # Test with DD_APPSEC_WAF_LOG_LEVEL (only latest go version)
           - go-version: '1.22'
             waf-log-level: TRACE
     name: ${{ toJSON(matrix) }}
--- a/.github/workflows/_test_bare_metal.yml
+++ b/.github/workflows/_test_bare_metal.yml
@@ -7,44 +7,12 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        runs-on: [ macos-14, macos-13, macos-12, ubuntu-22.04, ubuntu-20.04, windows-latest ]
-        go-version: [ "1.22.0-rc.2", "1.21", "1.20" ]
-        cgo-enabled: [ "0", "1" ] # test it compiles with and without cgo
-        go-tags:
-          - ''                      # Default behavior
-          - 'datadog.no_waf'        # Explicitly disabled WAF
-          - 'go1.23'                # Too recent go version (purego compatibility uncertain)
-          - 'appsec'                # Legacy build tag to enable appsec when cgo is disabled
-          - 'appsec,datadog.no_waf' # An awkward combination that must nevertheless compile
-          - 'datadog.no_waf,go1.23' # Explicitly disabled & too recent go version (purego compatibility uncertain)
-          - 'appsec,go1.23'         # Explicitly enabled w/o cgo & too recent go version (purego compatibility uncertain)
+        runs-on: [ macos-14, macos-13, macos-12, macos-11, ubuntu-22.04, ubuntu-20.04, ubuntu-latest-16-cores, windows-latest ]
+        go-version: [ "1.22.0", "1.21", "1.20", "1.19" ]
         include:
-          # gocheck2 is configured differently in go1.21 than in previous versions
-          - go-version: '1.21'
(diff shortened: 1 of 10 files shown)


.github/workflows/ci.sh
```diff
@@ -0,0 +1,59 @@
...
+# shellcheck disable=SC2046
+eval $(go env)
+
+case $GOVERSION in
+    *1.20*|*1.19* )
+        export GODEBUG=cgocheck=2
+        ;;
+    *)
+        export GOEXPERIMENT=cgocheck2
+esac
```
Julio-Guerra, line 16: This changes the behaviour so much that test must be also run without them being set up.

Here is the simple loop I did in my dd-trace-go CI PR https://github.com/DataDog/dd-trace-go/pull/2541/files#diff-dec1b363c4670835b0b6e96b3cebe6711e688d0dd14701a2cffaf6e6d1f60480R114-R124

.github/workflows/_test_containerized.yml
```diff
@@ -92,34 +62,17 @@ jobs:
...
-            -eDD_APPSEC_WAF_TIMEOUT="${{ matrix.waf-timeout || '5s' }}"         \
-            -eGODEBUG="${{ matrix.go-debug }}"                                  \
-            -eGOEXPERIMENT="${{ matrix.go-experiment }}"                        \
             -eGOMODCACHE="/go/pkg/mod"                                          \
             "${{ format(matrix.image, matrix.go-version) }}"
-      - name: Install alpine requirements
-        if: endsWith(matrix.image, '-alpine') && matrix.cgo-enabled == '1'
-        run: |-
-          docker exec -i gha-${{ github.run_id }}                               \
-            apk add gcc musl-dev libc6-compat
```
Julio-Guerra, line 107: I suggest you keep this here to do the same as the following al2 setup, and remove the setup out of ci.sh

Also, I find what I did here https://github.com/DataDog/dd-trace-go/pull/2541/files#diff-dec1b363c4670835b0b6e96b3cebe6711e688d0dd14701a2cffaf6e6d1f60480R226-R236 a bit simpler: everything in the same place instead of behind spread around multiple steps. It also simplifies the container creation and destruction.

Julio-Guerra (changes requested):

.github/workflows/_test_containerized.yml
eliottness: I can't put it back here because these packages needs to be installed when running with CGO and also specifically need to NOT be installed when running without CGO. Hence why, in my script, I run non-CGO stuff, then I install these packages, and finally I run the CGO stuff.

About what you did here ([DataDog/dd-trace-go#2541 (files)](https://github.com/DataDog/dd-trace-go/pull/2541/files#diff-dec1b363c4670835b0b6e96b3cebe6711e688d0dd14701a2cffaf6e6d1f60480R226-R236)) I feel that my script is a little too big to be directly embedded in yaml. Plus I would have to duplicate it since the same script is ran in the 2 different workflows (bare-metal & containerized)

.github/workflows/ci.sh
eliottness: done

.github/workflows/_test_containerized.yml
Julio-Guerra: Sounds good.
So you actually mix "smoke tests" with these.

.github/workflows/ci.sh
```diff
@@ -0,0 +1,66 @@
...
+run false datadog.no_waf,go1.23,appsec   # CGO disabled, WAF manually disabled, too recent go version with appsec explicitely enabled
+
+# Check if we are running on Alpine and install the required dependencies for cgo
+if [ -f /etc/os-release ] && grep -q Alpine < /etc/os-release; then
+  apk add gcc musl-dev libc6-compat
+fi
+
+run "$WAF_ENABLED" cgo                   # WAF enabled (but not on windows)
+run false go1.23,cgo                     # CGO disabled and too recent go version
+run false datadog.no_waf,cgo             # WAF manually disabled and CGO disabled
```
Julio-Guerra, line 65: "and CGO disable" but the cgo build tag is passed? 🤔

encoder.go
```diff
@@ -390,8 +390,8 @@ func (encoder *encoder) addTruncation(reason TruncationReason, size int) {
 // mesureObjectDepth traverses the provided object recursively to try and obtain
 // the real object depth, but limits itself to about 1ms of time budget, past
 // which it'll stop and return whatever it has go to so far.
-func (encoder *encoder) measureObjectDepth(obj reflect.Value) {
-	ctx, cancelCtx := context.WithTimeout(context.Background(), time.Millisecond)
+func (encoder *encoder) measureObjectDepth(obj reflect.Value, timeout time.Duration) {
+	ctx, cancelCtx := context.WithTimeout(context.Background(), timeout)
```
Julio-Guerra, line 394: I'd suggest managing 0 as disabled for test purposes - or use infinite in the tests?

eliottness: I am not really a fan of arbitrary values and this does not seem necessary to complexify this code just for tests. I will just put a huge value in the tests

eliottness: I put `time.Hour` because there is no `time.Infinity`. The test and job timeout will trigger before this anyway

.github/workflows/ci.sh
eliottness: Fixed

.github/workflows/_test_bare_metal.yml
```diff
@@ -7,44 +7,12 @@ jobs:
...
-        go-tags:
-          - ''                      # Default behavior
-          - 'datadog.no_waf'        # Explicitly disabled WAF
-          - 'go1.23'                # Too recent go version (purego compatibility uncertain)
-          - 'appsec'                # Legacy build tag to enable appsec when cgo is disabled
-          - 'appsec,datadog.no_waf' # An awkward combination that must nevertheless compile
-          - 'datadog.no_waf,go1.23' # Explicitly disabled & too recent go version (purego compatibility uncertain)
-          - 'appsec,go1.23'         # Explicitly enabled w/o cgo & too recent go version (purego compatibility uncertain)
+        runs-on: [ macos-14, macos-13, macos-12, macos-11, ubuntu-22.04, ubuntu-20.04, ubuntu-latest-16-cores, windows-latest ]
+        go-version: [ "1.22.0", "1.21", "1.20", "1.19" ]
```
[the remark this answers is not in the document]
eliottness: Fixed

Julio-Guerra (approved):

PR #75, opened 2024-02-27
feat: context-wide timeout

eliottness:
## Tasks

- [x] Add function `NewContextWithBudget`
- [x] Add new context method `Stats()` to get all time-related stats provided by go-libddwaf
- [x] The timeout value supplied in `Run()` is now a noop. 
- [x] Use the `timer` package 
- [x] The remaining time after the encoder is passed as value to ddwaf_run
- [x] The depth measuring will now happen only happen if there is still enough time to do it

## Reviewers

Please look at the Run function using the side-by-side view and not the inline diff view

### DIFF
--- a/context.go
+++ b/context.go
@@ -6,6 +6,8 @@
 package waf
 
 import (
+	stdErrors "errors"
+	"github.com/DataDog/go-libddwaf/v2/timer"
 	"sync"
 	"time"
 
@@ -25,13 +27,16 @@ type Context struct {
 	cgoRefs  cgoRefPool          // Used to retain go data referenced by WAF Objects the context holds
 	cContext bindings.WafContext // The C ddwaf_context pointer
 
-	// Stats
-	totalRuntimeNs        atomic.Uint64 // Cumulative internal WAF run time - in nanoseconds - for this context.
-	totalOverallRuntimeNs atomic.Uint64 // Cumulative overall run time - in nanoseconds - for this context.
-	timeoutCount          atomic.Uint64 // Cumulative timeout count for this context.
+	timeoutCount atomic.Uint64 // Cumulative timeout count for this context.
 
 	// Mutex protecting the use of cContext which is not thread-safe and cgoRefs.
 	mutex sync.Mutex
+
+	// timer registers the time spent in the WAF and go-libddwaf
+	timer timer.NodeTimer
+
+	// stats stores the cumulative time spent in various parts of the WAF
+	stats metrics
 }
 
 // NewContext returns a new WAF context of to the given WAF handle.
@@ -40,6 +45,15 @@ type Context struct {
 // handle. A nil value is returned when the WAF handle can no longer be used
 // or the WAF context couldn't be created.
 func NewContext(handle *Handle) *Context {
+	return NewContextWithBudget(handle, timer.UnlimitedBudget)
+}
+
+// NewContextWithBudget returns a new WAF context of to the given WAF handle.
+// A nil value is returned when the WAF handle was released or when the
+// WAF context couldn't be created.
+// handle. A nil value is returned when the WAF handle can no longer be used
+// or the WAF context couldn't be created.
+func NewContextWithBudget(handle *Handle, budget time.Duration) *Context {
 	// Handle has been released
 	if !handle.retain() {
 		return nil
@@ -51,7 +65,12 @@ func NewContext(handle *Handle) *Context {
 		return nil
 	}
 
-	return &Context{handle: handle, cContext: cContext}
+	timer, err := timer.NewTreeTimer(timer.WithBudget(budget), timer.WithComponents("dd.appsec.waf.run"))
+	if err != nil {
+		return nil
+	}
+
+	return &Context{handle: handle, cContext: cContext, timer: timer, stats: metrics{data: make(map[string]time.Duration, 6)}}
 }
 
 // RunAddressData provides address data to the Context.Run method. If a given key is present in both
@@ -75,45 +94,74 @@ func (d RunAddressData) isEmpty() bool {
 // matches and actions can still be returned, for instance in the case of a timeout error. Errors can be tested against
 // the RunError type.
 // Struct fields having the tag `ddwaf:"ignore"` will not be encoded and sent to the WAF
-func (context *Context) Run(addressData RunAddressData, timeout time.Duration) (res Result, err error) {
+// if the output of TotalTime() exceeds the value of Timeout, the function will immediately return with an error
+func (context *Context) Run(addressData RunAddressData, budget time.Duration) (res Result, err error) {
 	if addressData.isEmpty() {
 		return
 	}
 
-	now := time.Now()
 	defer func() {
-		dt := time.Since(now)
-		context.totalOverallRuntimeNs.Add(uint64(dt.Nanoseconds()))
+		if stdErrors.Is(err, errors.ErrTimeout) {
+			context.timeoutCount.Inc()
+		}
 	}()
 
-	// At this point, the only error we can get is an error in case the top level object is a nil map, but this
-	// behaviour is expected since either persistent or ephemeral addresses are allowed to be null one at a time.
-	// In this case, EncodeAddresses will return nil contrary to Encode which will return an nil wafObject,
-	// which is what we need to send to ddwaf_run to signal that the address data is empty.
-	var persistentData *bindings.WafObject = nil
-	var ephemeralData *bindings.WafObject = nil
-	persistentEncoder := newLimitedEncoder()
-	ephemeralEncoder := newLimitedEncoder()
-	if addressData.Persistent != nil {
-		persistentData, _ = persistentEncoder.EncodeAddresses(addressData.Persistent)
+	// If the context has already timed out, we don't need to run the WAF again
+	if context.timer.SumExhausted() {
+		return Result{}, errors.ErrTimeout
 	}
 
-	if addressData.Ephemeral != nil {
-		ephemeralData, _ = ephemeralEncoder.EncodeAddresses(addressData.Ephemeral)
-
+	runTimer, err := context.timer.NewNode("dd.appsec.waf.run", timer.WithBudget(budget),
+		timer.WithComponents(
+			"dd.appsec.waf.encode.persistent",
+			"dd.appsec.waf.encode.ephemeral",
+			"dd.appsec.waf.lock",
+			"dd.appsec.waf.duration_ext",
+			"dd.appsec.waf.duration",
+		),
+	)
+	if err != nil {
+		return Result{}, err
 	}
-	// The WAF releases ephemeral address data at the end of each run call, so we need not keep the Go values live beyond
+
+	runTimer.Start()
+	defer func() {
+		context.stats.add("dd.appsec.waf.run", runTimer.Stop())
+		context.stats.append(runTimer.Stats())
+	}()
+
+	persistentData, persistentEncoder, err := context.encodeOneAddressType(addressData.Persistent, runTimer.MustLeaf("dd.appsec.waf.encode.persistent"))
+	if err != nil {
+		return res, err
+	}
+
+	// The WAF releases ephemeral address data at the max of each run call, so we need not keep the Go values live beyond
 	// that in the same way we need for persistent data. We hence use a separate encoder.
+	ephemeralData, ephemeralEncoder, err := context.encodeOneAddressType(addressData.Ephemeral, runTimer.MustLeaf("dd.appsec.waf.encode.persistent"))
+	if err != nil {
+		return res, err
+	}
 
 	// ddwaf_run cannot run concurrently and the next append write on the context state so we need a mutex
+	waitTimer := runTimer.MustLeaf("dd.appsec.waf.lock")
+	waitTimer.Start()
 	context.mutex.Lock()
+	waitTimer.Stop()
+
 	defer context.mutex.Unlock()
+
+	if runTimer.SumExhausted() {
+		return res, errors.ErrTimeout
+	}
 
 	// Save the Go pointer references to addressesToData that were referenced by the encoder
 	// into C ddwaf_objects. libddwaf's API requires to keep this data for the lifetime of the ddwaf_context.
 	defer context.cgoRefs.append(persistentEncoder.cgoRefs)
 
-	res, err = context.run(persistentData, ephemeralData, timeout)
+	wafExtTimer := runTimer.MustLeaf("dd.appsec.waf.duration_ext")
+	res, err = context.run(persistentData, ephemeralData, wafExtTimer, runTimer.SumRemaining())
+
+	runTimer.AddTime("dd.appsec.waf.duration", res.TimeSpent)
 
 	// Ensure the ephemerals don't get optimized away by the compiler before the WAF had a chance to use them.
 	unsafe.KeepAlive(ephemeralEncoder.cgoRefs)
@@ -157,21 +205,38 @@ func merge[K comparable, V any](a, b map[K][]V) (merged map[K][]V) {
 	return
 }
 
+// encodeOneAddressType encodes the given addressData values and returns the corresponding WAF object and its refs.
+// If the addressData is empty, it returns nil for the WAF object and an empty ref pool.
+// At this point, the only error we can get is an error in case the top level object is a nil map, but this
+// behaviour is expected since either persistent or ephemeral addresses are allowed to be null one at a time.
+// In this case, EncodeAddresses will return nil contrary to Encode which will return an nil wafObject,
+// which is what we need to send to ddwaf_run to signal that the address data is empty.
+func (context *Context) encodeOneAddressType(addressData map[string]any, timer timer.Timer) (*bindings.WafObject, encoder, error) {
+	encoder := newLimitedEncoder(timer)
+	if addressData == nil {
+		return nil, encoder, nil
+	}
+
+	data, _ := encoder.EncodeAddresses(addressData)
+
+	if timer.Exhausted() {
+		return nil, encoder, errors.ErrTimeout
+	}
+
+	return data, encoder, nil
+}
+
 // run executes the ddwaf_run call with the provided data on this context. The caller is responsible for locking the
 // context appropriately around this call.
-func (context *Context) run(persistentData, ephemeralData *bindings.WafObject, timeout time.Duration) (Result, error) {
+func (context *Context) run(persistentData, ephemeralData *bindings.WafObject, timer timer.Timer, timeBudget time.Duration) (Result, error) {
 	result := new(bindings.WafResult)
 	defer wafLib.WafResultFree(result)
 
-	ret := wafLib.WafRun(context.cContext, persistentData, ephemeralData, result, uint64(timeout/time.Microsecond))
+	timer.Start()
+	defer timer.Stop()
+	ret := wafLib.WafRun(context.cContext, persistentData, ephemeralData, result, uint64(timeBudget.Microseconds()))
 
-	context.totalRuntimeNs.Add(result.TotalRuntime)
-	res, err := unwrapWafResult(ret, result)
-	if err == errors.ErrTimeout {
-		context.timeoutCount.Inc()
-	}
-
-	return res, err
+	return unwrapWafResult(ret, result)
 }
 
 func unwrapWafResult(ret bindings.WafReturnCode, result *bindings.WafResult) (res Result, err error) {
@@ -204,6 +269,7 @@ func unwrapWafResult(ret bindings.WafReturnCode, result *bindings.WafResult) (re
 		}
 	}
 
+	res.TimeSpent = time.Duration(result.TotalRuntime)
 	return res, err
 }
 
@@ -216,7 +282,7 @@ func (context *Context) Close() {
 	defer context.mutex.Unlock()
 
 	wafLib.WafContextDestroy(context.cContext)
-	unsafe.KeepAlive(context.cgoRefs) // Keep the Go pointer references until the end of the context
+	unsafe.KeepAlive(context.cgoRefs) // Keep the Go pointer references until the max of the context
 	defer context.handle.release()    // Reduce the reference counter of the Handle.
 
 	context.cgoRefs = cgoRefPool{} // The data in context.cgoRefs is no longer needed, explicitly release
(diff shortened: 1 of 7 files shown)


context.go
```diff
@@ -75,52 +77,69 @@ func (d RunAddressData) isEmpty() bool {
...
+		return res, err
 	}
-	// The WAF releases ephemeral address data at the end of each run call, so we need not keep the Go values live beyond
+
+	// The WAF releases ephemeral address data at the max of each run call, so we need not keep the Go values live beyond
 	// that in the same way we need for persistent data. We hence use a separate encoder.
+	ephemeralData, ephemeralEncoder, err := context.encodeOneAddressType(addressData.Ephemeral, &runTimings)
+	if err != nil {
+		return res, err
+	}
```
RomainMuller, line 117: If we time out during encode, but we successfully encoded _something_, are we returning an error, or are we proceeding with partial input?

I feel like proceeding with partial input might be preferable (we spent time encoding those things, this is pure sunk cost unless we actually use that).

If we process partial input, it might be worth wondering if we should encode persistent or ephemeral data first?

context.go
```diff
@@ -75,52 +77,69 @@ func (d RunAddressData) isEmpty() bool {
...
+		res.Truncations = merge(persistentEncoder.Truncations(), ephemeralEncoder.Truncations())
+	}()
 
 	// ddwaf_run cannot run concurrently and the next append write on the context state so we need a mutex
-	context.mutex.Lock()
-	defer context.mutex.Unlock()
+	// waitingNs is only indicative and is not taken into account when calculating the global timeout
+	stopwatch.Timed(&runTimings.WaitingNs, func() {
+		context.mutex.Lock()
+	})
```
RomainMuller: Imagine:

```go
runTimings.TimedWait(context.mutex.Lock)
```

context.go
```diff
@@ -75,52 +77,69 @@ func (d RunAddressData) isEmpty() bool {
...
-	// Save the Go pointer references to addressesToData that were referenced by the encoder
-	// into C ddwaf_objects. libddwaf's API requires to keep this data for the lifetime of the ddwaf_context.
-	defer context.cgoRefs.append(persistentEncoder.cgoRefs)
+	defer context.mutex.Unlock()
 
-	res, err = context.run(persistentData, ephemeralData, timeout)
+	// Timer for the actual run call
+	stopwatch.Timed(&runTimings.ExtRuntimeNs, func() {
+		res, err = context.run(persistentData, ephemeralData, runTimings.Remaining())
+	})
```
RomainMuller: Following with the previous train of thought:
```go
runTimings.TimedExternal(func(){
  res, err = context.run(persistentData, ephemeralData, runTimings.Remaining())
})
```

context.go
```diff
@@ -157,21 +176,37 @@ func merge[K comparable, V any](a, b map[K][]V) (merged map[K][]V) {
 	return
 }
 
+// encodeOneAddressType encodes the given addressData values and returns the corresponding WAF object and its refs.
+// If the addressData is empty, it returns nil for the WAF object and an empty ref pool.
+// At this point, the only error we can get is an error in case the top level object is a nil map, but this
+// behaviour is expected since either persistent or ephemeral addresses are allowed to be null one at a time.
+// In this case, EncodeAddresses will return nil contrary to Encode which will return an nil wafObject,
+// which is what we need to send to ddwaf_run to signal that the address data is empty.
+func (context *Context) encodeOneAddressType(addressData map[string]any, runTimings *Timings) (*bindings.WafObject, *encoder, error) {
```
RomainMuller: (Optional) I'd have called this `encodeAddressData[Map]`

context.go
```diff
@@ -157,21 +176,37 @@ func merge[K comparable, V any](a, b map[K][]V) (merged map[K][]V) {
...
+	var data *bindings.WafObject = nil
+	encoder := newLimitedEncoder(runTimings.Remaining())
```
RomainMuller, line 188: This appears to extract the `Remaining` from the `runTimings`, which means the encoder might be operating based on a stale deadline?

context.go
```diff
@@ -157,21 +176,37 @@ func merge[K comparable, V any](a, b map[K][]V) (merged map[K][]V) {
...
+	ret := wafLib.WafRun(context.cContext, persistentData, ephemeralData, result, uint64(timeout.Microseconds()))
 
-	context.totalRuntimeNs.Add(result.TotalRuntime)
-	res, err := unwrapWafResult(ret, result)
-	if err == errors.ErrTimeout {
-		context.timeoutCount.Inc()
-	}
-
-	return res, err
+	context.timings.RuntimeNs.Add(int64(result.TotalRuntime))
```
RomainMuller, line 209: `uint64` to `int64` conversion feels gross (the other way around is "fine" if you can guarantee it's never negative).

context.go
```diff
@@ -216,7 +251,7 @@ func (context *Context) Close() {
 	defer context.mutex.Unlock()
 
 	wafLib.WafContextDestroy(context.cContext)
-	unsafe.KeepAlive(context.cgoRefs) // Keep the Go pointer references until the end of the context
+	unsafe.KeepAlive(context.cgoRefs) // Keep the Go pointer references until the max of the context
```
RomainMuller, line 255: "The max of the context" does not mean much here...

Maybe:
```suggestion
	unsafe.KeepAlive(context.cgoRefs) // Keep the Go pointer references alive as long as the context is usable
```

context.go
```diff
@@ -225,11 +260,29 @@ func (context *Context) Close() {
...
 }
 
 // TotalTimeouts returns the cumulated amount of WAF timeouts across various run calls within the same WAF context.
+// Deprecated: use Timings instead
 func (context *Context) TotalTimeouts() uint64 {
-	return context.timeoutCount.Load()
+	return context.timeBudgetExpirationCount.Load()
+}
+
+// Timings returns the cumulative time spent in various parts of the WAF, all in nanoseconds
```
RomainMuller, line 275: Same as before, the units is a concern of `Timings`, not of this function.

```suggestion
// Timings returns the cumulative time spent in various parts of the WAF
```

context.go
```diff
@@ -225,11 +260,29 @@ func (context *Context) Close() {
...
+// and the timeout value used
+func (context *Context) Timings() Timings {
+	return context.timings
+}
+
+// SetTimeBudget sets the maximum time the WAF is allowed to run for the given context cumulatively on all run calls.
+func (context *Context) SetTimeBudget(timeBudget time.Duration) {
+	context.timings.TimeBudget = timeBudget
+}
```
RomainMuller, line 284: What is the use-case for being able to change the `TimeBudget` after the context has been created? It seems to me this field is expected to be immutable?

encoder.go
```diff
@@ -61,8 +65,9 @@ type native interface {
 	int64 | uint64 | uintptr
 }
 
-func newLimitedEncoder() encoder {
+func newLimitedEncoder(timeout time.Duration) encoder {
```
RomainMuller, line 68: (nitpick) This is actually a budget, not a timeout.

encoder.go
```diff
@@ -61,8 +65,9 @@ type native interface {
...
 	return encoder{
+		watch:            stopwatch.NewTimedStopwatch(timeout),
```
RomainMuller, line 70: I'll get to that later... but see how this stutters:

```go
  stopwatch.NewTimedStopwatch(timeout)
//~~~~~~~~~         ~~~~~~~~~
```

encoder.go
```diff
@@ -85,9 +91,12 @@ func (encoder *encoder) Encode(data any) (wo *bindings.WafObject, err error) {
 	value := reflect.ValueOf(data)
 	wo = &bindings.WafObject{}
 
-	err = encoder.encode(value, wo, encoder.objectMaxDepth)
-	if len(encoder.truncations[ObjectTooDeep]) != 0 {
-		encoder.measureObjectDepth(value, time.Millisecond)
+	stopwatch.TimedByStopwatch(encoder.watch, func() {
```
RomainMuller, line 94: More stutter!

```go
  stopwatch.TimedByStopwatch(encoder.watch, func() {
//~~~~~~~~~        ~~~~~~~~~
```

encoder.go
```diff
@@ -85,9 +91,12 @@ func (encoder *encoder) Encode(data any) (wo *bindings.WafObject, err error) {
...
+		err = encoder.encode(value, wo, encoder.objectMaxDepth)
+	})
+
+	if len(encoder.truncations[ObjectTooDeep]) != 0 && !encoder.watch.Expired() {
+		encoder.measureObjectDepth(value, encoder.watch.Remaining())
 	}
```
RomainMuller, line 100: MAYBE we wanna do this even if the watch expired? (but also maybe not)

The `measureObjectDepth` function caps itself to a 1ms time budget.

internal/stopwatch/stopwatch.go
```diff
@@ -0,0 +1,175 @@
...
+package stopwatch
+
+import (
+	"sync"
+	"sync/atomic"
+	"time"
+)
+
+// Stopwatch is a simple interface to measure time.
+// Stopwatch implementations are not thread-safe.
```
RomainMuller, line 15:
```suggestion
// Stopwatch provides methods useful to measure time spent on one or more
// tasks, and to limit the overall time consumed by these tasks.
// Implementations are not required to be thread-safe.
```

context.go
```diff
@@ -75,52 +77,69 @@ func (d RunAddressData) isEmpty() bool {
...
+		return res, errors.ErrTimeout
 	}
 
-	if addressData.Ephemeral != nil {
-		ephemeralData, _ = ephemeralEncoder.EncodeAddresses(addressData.Ephemeral)
+	runTimings := context.timings.newRun()
+	startTime := time.Now()
+	defer func() {
+		runTimings.RunNs.Add(time.Since(startTime).Nanoseconds())
+		context.timings.absorb(&runTimings)
```
Hellzy, line 103: I think this misbehaves if `context.Run()` gets called more than once per context.
`timings.newRun()` creates a `Timings` where the run duration is the current value of context.timings.
This works fine when no run has occured, but if one has already occurred and the value is not zero then `absorb` will end up adding too much to the context timings run duration (it will basically do context run duration * 2 + run duration) .

internal/stopwatch/stopwatch.go
```diff
@@ -0,0 +1,175 @@
...
+type Stopwatch interface {
+	// Start starts the stopwatch and returns the start time.
+	Start() time.Time
```
RomainMuller, line 18: What happens if the stopwatch was already started?

internal/stopwatch/stopwatch.go
```diff
@@ -0,0 +1,175 @@
...
+	// Remaining returns the remaining time before the stopwatch expires.
+	Remaining() time.Duration
+	// Expired returns true if the stopwatch has expired.
+	Expired() bool
```
RomainMuller, line 22:
```suggestion
	// Exhausted returns true if the stopwatch has exhausted its configured budget.
	Exhausted() bool
```

The documentation may also explain what `Exhausted` means in terms of `Remaining` (is 0) and `Spent`/`Consumed` (is == configured budget)

internal/stopwatch/stopwatch.go
```diff
@@ -0,0 +1,175 @@
...
+	// Stop stops the stopwatch and returns the time spent.
+	Stop() time.Duration
```
RomainMuller, line 24: What happens if it was already stopped? Is this a definitive operation? Can it be started again after it's been stopped (then, is this really `Pause`?)

internal/stopwatch/stopwatch.go
```diff
@@ -0,0 +1,175 @@
...
+	// Spent returns the time spent on the stopwatch.
+	Spent() time.Duration
```
RomainMuller, line 26:
```suggestion
	// Consumed returns the time consumed by tasks measured by this stopwatch.
	Consumed() time.Duration
```

Hellzy (changes requested):
I'm concerned about the way we use `timings.newRun()` combined with `timings.absorb()` in case of several WAF runs per context. Can you check?

internal/stopwatch/stopwatch.go
```diff
@@ -0,0 +1,175 @@
...
+
+import (
+	"sync"
+	"sync/atomic"
+	"time"
+)
+
+// Stopwatch is a simple interface to measure time.
+// Stopwatch implementations are not thread-safe.
+type Stopwatch interface {
```
RomainMuller, line 16: I **strongly** recommend sorting `Stopwatch` methods by "type of use":
- Control whether it's counting or not:
  - `Start`
  - `Stop`
- Inspect resource consumption
  - `Remaining` 
  - `Spent`
  - `Expired`

internal/stopwatch/stopwatch.go
```diff
@@ -0,0 +1,175 @@
...
+	Spent() time.Duration
+}
+
+type unlimitedStopwatch struct {
+	start time.Time
+	end   time.Time
+}
+
+// NewUnlimitedStopwatch returns a new unlimited stopwatch with no expiration.
+func NewUnlimitedStopwatch() Stopwatch {
```
RomainMuller, line 35: We're in a `stopwatch` package and should reduce stutter on the call sites:

```suggestion
// Unlimited returns a new stopwatch with infinite time budget.
func Unlimited() Stopwatch {
```

internal/stopwatch/stopwatch.go
```diff
@@ -0,0 +1,175 @@
...
+	return &unlimitedStopwatch{}
+}
+
+func (stopwatch *unlimitedStopwatch) Start() time.Time {
+	stopwatch.start = time.Now()
+	return stopwatch.start
+}
```
RomainMuller, line 42: This allows calling `Start` multiple times in a row, with probably unexpected results (in other words, this is very easy to mis-use).

internal/stopwatch/stopwatch.go
```diff
@@ -0,0 +1,175 @@
...
+
+func (stopwatch *unlimitedStopwatch) Remaining() time.Duration {
+	return 0
+}
```
RomainMuller, line 46: This should return `time.Duration(math.MaxInt64)` in my opinion.

internal/stopwatch/stopwatch.go
```diff
@@ -0,0 +1,175 @@
...
+
+func (stopwatch *unlimitedStopwatch) Expired() bool {
+	return false
+}
+
+func (stopwatch *unlimitedStopwatch) Stop() time.Duration {
+	stopwatch.end = time.Now()
+	return stopwatch.Spent()
+}
```
RomainMuller, line 55: Same as above - this looks way too easy to mis-use.

internal/stopwatch/stopwatch.go
```diff
@@ -0,0 +1,175 @@
...
+
+func (stopwatch *unlimitedStopwatch) Spent() time.Duration {
+	return stopwatch.end.Sub(stopwatch.start)
+}
+
+type timedStopwatch struct {
```
RomainMuller, line 61: (Optional) I would have the "unlimited" and "limited" stopwatch implementations in their own individual files.

Also, I think `timed` isn't a great name here... This is `budgeted` or `limited`.

internal/stopwatch/stopwatch.go
```diff
@@ -0,0 +1,175 @@
...
+	start    time.Time
+	duration time.Duration
+	end      time.Time
+	max      time.Time
```
RomainMuller, line 65: Would love some commentary to indicate what each field represents.

Also, `max` perhaps should be called `cap` or `limit` or `deadline`.

internal/stopwatch/stopwatch.go
```diff
@@ -0,0 +1,175 @@
...
+}
+
+// NewTimedStopwatch returns a new stopwatch that expires after the given duration.
+func NewTimedStopwatch(duration time.Duration) Stopwatch {
```
RomainMuller, line 69:
```suggestion
// WithBudget returns a new stopwatch that expires after the given duration.
func WithBudget(budget time.Duration) Stopwatch {
```

internal/stopwatch/stopwatch.go
```diff
@@ -0,0 +1,175 @@
...
+	return &timedStopwatch{
+		duration: duration,
+	}
+}
+
+func (stopwatch *timedStopwatch) Start() time.Time {
+	stopwatch.start = time.Now()
+	stopwatch.max = stopwatch.start.Add(stopwatch.duration)
+	return stopwatch.start
+}
```
RomainMuller, line 79: Same as before, this looks too easy to mis-use (the same applies to `Stop` here).
Calling `Start` effectively resets the budget counter...

internal/stopwatch/stopwatch.go
```diff
@@ -0,0 +1,175 @@
...
+
+func (stopwatch *timedStopwatch) Remaining() time.Duration {
+	return time.Until(stopwatch.max)
```
RomainMuller, line 82: This is going to return incorrect if `Start` has never been called.

internal/stopwatch/stopwatch.go
```diff
@@ -0,0 +1,175 @@
...
+		ongoing     int32
+		oldestStart time.Duration
+		duration    time.Duration
+	}
+
+	LocalStopWatch struct {
+		s           *SharedStopWatch
+		start, stop time.Duration
+	}
+)
```
RomainMuller, line 111: These should also live in their own file.

internal/stopwatch/stopwatch.go
```diff
@@ -0,0 +1,175 @@
...
+
+func NewSharedStopWatch() *SharedStopWatch {
```
RomainMuller, line 113: Documentation?

```suggestion
func Shared() *SharedStopWatch {
```

internal/stopwatch/stopwatch.go
```diff
@@ -0,0 +1,175 @@
...
+}
+
+func (s *SharedStopWatch) Start() (ls LocalStopWatch) {
+	ls = LocalStopWatch{
+		s:     s,
+		start: time.Since(s.zeroT),
+		stop:  -1,
+	}
+
+	if atomic.AddInt32(&s.ongoing, 1) > 1 {
```
RomainMuller, line 127: You shouldn't use `atomic` functions... Only the types... The functions have alignment requirements that you're responsible to ensure of yourself; this is fiddly. Also, this pattern makes mis-use way too easy.

internal/stopwatch/stopwatch.go
```diff
@@ -0,0 +1,175 @@
...
+		return
+	}
+
+	// Exclusively lock the watch to enforce the coherency of oldestStart's value
```
RomainMuller, line 131:
```suggestion
	// Exclusively lock the watch to enforce the coherence of oldestStart's value
```

internal/stopwatch/timed.go
```diff
@@ -0,0 +1,25 @@
...
+
+package stopwatch
+
+import (
+	"go.uber.org/atomic"
+	"time"
+)
+
+// Timed runs the given function and adds the time it took to the given atomic.Int64. Thread-safe
+func Timed(timer *atomic.Int64, f func()) {
```
RomainMuller: This is a place where you want to be specific that the timer is nanos... This function being exported is also a bit weird... especially since this package provides the `Stopwatch` which is supposed to make this redundant?

internal/stopwatch/timed.go
```diff
@@ -0,0 +1,25 @@
...
+	start := time.Now()
+	f()
+	timer.Add(time.Since(start).Nanoseconds())
+}
+
+// TimedByStopwatch runs the given function and adds the time it took to the given stopwatch. NOT Thread-safe
+func TimedByStopwatch(watch Stopwatch, f func()) {
```
RomainMuller: This should 100% be a member of `Stopwatch` IMHO. You could have a design such as:

- `Stopwatch` has this `TimeFunc` method + embeds `stopwatchBase` interface that is your current `Stopwatch` methods
- A `stopwatch` struct embeds a `stopwatchBase` value + implements `TimeFunc`
- Voilà.

internal/stopwatch/timed.go
```diff
@@ -0,0 +1,25 @@
...
+	watch.Start()
+	f()
+	watch.Stop()
```
RomainMuller: This isn't `panic` safe, BTW... you should:

```suggestion
	watch.Start()
	defer watch.Stop()
	f()
```

timings.go
```diff
@@ -0,0 +1,68 @@
...
+	WaitingNs atomic.Int64 // Cumulative time spent waiting for the WAF to run for this context.
+
+	// RunNs is the cumulative time spent running in waf.Run() for this context.
+	// this value should vaguely be superior to WaitingNs + EncoderNs + ExtRuntimeNs
+	RunNs atomic.Int64 // Cumulative time spent running in go-libddwaf for this context.
+
+	// TimeBudget is the maximum time the WAF is allowed to run for the given context cumulatively on all run calls.
+	// this value is not expected to change after the first run call and is not thread-safe.
+	TimeBudget time.Duration
+}
```
RomainMuller, line 41:
```suggestion
type Timings struct {
	// RuntimeNs is the cumulative internal WAF run time for this context.
	// It corresponds to the value returned by in `ddwaf_result`
	RuntimeNs atomic.Int64

	// ExtRuntimeNs is the cumulative external WAF run time for this context.
	// It is a manual timer that is started when the WAF is called and stopped when the WAF returns including
	// go native code call switch overhead.
	// This value should always be strictly superior to RuntimeNs
	ExtRuntimeNs atomic.Int64

	// EncoderNs is the cumulative time spent encoding addresses for this context.
	// it correspond to the value of Stopwatch.Spent() for both persistent and ephemeral encoders
	EncoderNs atomic.Int64

	// WaitingNs is the cumulative time spent waiting for the WAF to run for this context.
	// It's the time spend waiting for Context.mutex to be unlocked.
	WaitingNs atomic.Int64

	// RunNs is the cumulative time spent running in waf.Run() for this context.
	// this value should vaguely be superior to WaitingNs + EncoderNs + ExtRuntimeNs
	RunNs atomic.Int64

	// TimeBudget is the maximum time the WAF is allowed to run for the given context cumulatively on all run calls.
	// this value is not expected to change after the first run call and is not thread-safe.
	TimeBudget time.Duration
}
```

timings.go
```diff
@@ -0,0 +1,68 @@
...
+
+func (timings *Timings) absorb(other *Timings) {
```
RomainMuller: `merge`? `add`?

RomainMuller (changes requested):

context.go
```diff
@@ -75,45 +93,70 @@ func (d RunAddressData) isEmpty() bool {
...
+
+	// The WAF releases ephemeral address data at the max of each run call, so we need not keep the Go values live beyond
 	// that in the same way we need for persistent data. We hence use a separate encoder.
+	ephemeralData, ephemeralEncoder, err := encodeOneAddressType(addressData.Ephemeral, runTimer.MustLeaf("dd.appsec.waf.encode.ephemeral"))
+	if err != nil {
+		return res, err
+	}
 
-	// ddwaf_run cannot run concurrently and the next append write on the context state so we need a mutex
+	// ddwaf_run cannot run concurrently and the next merge write on the context state so we need a mutex
```
RomainMuller, line 144: That append is the one done in the `defer` above? If so, it'll happen _after_ the lock is released... which this comment would imply is not fine?

context.go
```diff
@@ -157,21 +200,41 @@ func merge[K comparable, V any](a, b map[K][]V) (merged map[K][]V) {
...
 	defer wafLib.WafResultFree(result)
 
-	ret := wafLib.WafRun(context.cContext, persistentData, ephemeralData, result, uint64(timeout/time.Microsecond))
-
-	context.totalRuntimeNs.Add(result.TotalRuntime)
-	res, err := unwrapWafResult(ret, result)
-	if err == errors.ErrTimeout {
-		context.timeoutCount.Inc()
+	if timeBudget == timer.UnlimitedBudget {
+		timeBudget = timer.UnlimitedBudget / 1000 // Value is too big for the WAF
```
RomainMuller, line 230: How so? Are we not wiring an `int64` all the way? (I'm a bit surprised, but this does not necessarily need changing... maybe just clarify this statement?)

context.go
```diff
@@ -225,11 +289,59 @@ func (context *Context) Close() {
...
+
+type metrics struct {
+	data  map[string]time.Duration
+	mutex sync.RWMutex
+}
+
+func (metrics *metrics) add(key string, duration time.Duration) {
+	metrics.mutex.Lock()
+	defer metrics.mutex.Unlock()
+	metrics.data[key] += duration
```
RomainMuller, line 317: I don't see a constructor here, and it appears to me that this write on the `map[string]time.Duration` is susceptible to attempt writing on `nil` as a result?

If this is expected to always be constructed with a non-`nil` map, then I'd rather have initialization funelled through a constructor function

context.go
```diff
@@ -225,11 +289,59 @@ func (context *Context) Close() {
...
+// merge merges the current metrics with the new run
+func (metrics *metrics) merge(other map[string]time.Duration) {
+	metrics.mutex.Lock()
+	defer metrics.mutex.Unlock()
+	for key, val := range other {
+		prev, ok := metrics.data[key]
+		if !ok {
+			prev = 0
+		}
+		metrics.data[key] = prev + val
```
RomainMuller, line 345: It feels like this could be delegated to `metrics.add`?

```suggestion
		metrics.add(key, val)
```

RomainMuller (approved):
Just a couple mostly cosmetic nits... looks good now!

context.go
```diff
@@ -51,7 +64,12 @@ func NewContext(handle *Handle) *Context {
 		return nil
 	}
 
-	return &Context{handle: handle, cContext: cContext}
+	timer, err := timer.NewTreeTimer(timer.WithBudget(budget), timer.WithComponents("_dd.appsec.waf.run"))
+	if err != nil {
```
Hellzy, line 68: It kinda bugs me that we're dropping errors like this...
But since we don't really log things in this repo the alternative would be to change the API to return the error, meaning a new major V.
I guess we can put a pin in that for now but I think it's worth keeping in mind for when we plan a new major release. WDYT?

Hellzy (commented):
SGTM, would you mind ordering the new imports like we usually tend to do? (std vs local vs ext)

context.go
eliottness: We are starting to get a good list of things to do for the v3. I just added a task in jira to track [that](https://datadoghq.atlassian.net/browse/APPSEC-52238?atlOrigin=eyJpIjoiMGI0NThkOWFiODJhNDVmYzg1MmI5N2FlYjliYTQ5NjMiLCJwIjoiaiJ9)

PR #78, opened 2024-03-05
chore: also run containerized tests on ARM

RomainMuller:
Also removed 1.19 suite since the package's `go.mod` declares `go 1.20`...

### DIFF
--- a/.github/workflows/_test_containerized.yml
+++ b/.github/workflows/_test_containerized.yml
@@ -7,6 +7,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        arch: [ amd64, arm64 ]
         image:
           # Standard golang images
           - golang:{0}-alpine
@@ -15,7 +16,7 @@ jobs:
           - golang:{0}-buster
           # RPM-based image
           - amazonlinux:2 # pretty popular on AWS workloads
-        go-version: [ "1.22", "1.21", "1.20", "1.19" ]
+        go-version: [ "1.22", "1.21", "1.20" ]
         include:
           # Test with DD_APPSEC_WAF_LOG_LEVEL (only latest go, without any particular tag)
           - go-version: '1.22'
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -14,10 +14,13 @@ on:
 
 jobs:
   bare-metal:
+    name: Directly on GitHub Runner
```
eliottness, line 17:
```suggestion
    name: GitHub Runner
```

.github/workflows/_test_containerized.yml
```diff
@@ -28,15 +29,29 @@ jobs:
...
-    runs-on: ubuntu-latest
+    name: ${{ matrix.arch }} ${{ format(matrix.image, matrix.go-version) }} go${{ matrix.go-version }}${{ matrix.waf-log-level && format(' (DD_APPSEC_WAF_LOG_LEVEL={0})', matrix.waf-log-level) || '' }}
+    # We use ARM runners when needed to avoid the performance hit of QEMU
+    runs-on: ${{ matrix.arch == 'amd64' && 'ubuntu-latest' || 'arm-4core-linux' }}
     steps:
+      # Docker is not present on early-access ARM runners
+      - name: Prepare ARM Runner
+        if: matrix.arch == 'arm64'
+        run: |-
+          for pkg in docker.io docker-doc docker-compose docker-compose-v2 podman-docker containerd runc; do sudo apt-get remove -y $pkg || echo "Not present: $pkg"; done
```
eliottness, line 44: Isn't there an apt action ? (that would support caching for us)

eliottness (approved):

.github/workflows/_test_containerized.yml
RomainMuller: The early-access ARM runners don't have a full fledged image and lack a lot of tools needed by some actions, so even if there was one (I don't know about one), it's not certain it'd work there...

PR #79, opened 2024-03-07
feat: setup timer package

eliottness:
## What

This PR adds the new package `timer` that adds 2 new interfaces: `Timer` and `NodeTimer` that can be used to create trees of timers and use them to better dispatch the time of each request the customer allocated to ASM.

## Motivation

The current way to time this package and, more generally, appsec, is becoming a blocker to customers and Dogfooding. Having a better way to manage the budget we are allowed to get from the current application appsec is running on is crucial to make this time worth it.

## Note to reviewers

The first read should be the `timer.go` file where interfaces sits.

### DIFF
--- a/timer/base_timer.go
+++ b/timer/base_timer.go
@@ -18,18 +18,18 @@ type baseTimer struct {
 	// start is the time when the timer was started
 	start time.Time
 
-	// spent is the time spent on the timer, set after calling stop
-	spent time.Duration
-
-	// config is the configuration of the timer
-	config Config
+	// parent is the parent timer. It is used to progate the stop of the timer to the parent timer and get the remaining time in case the budget has to be inherited.
+	parent NodeTimer
 
 	// parentComponent is the component stored in the map of the parent timer. We keep a reference to it to update its spent time
 	// once the timer is stopped. It may be nil if the timer is not a child of a tree timer, like when calling NewTimer.
 	parentComponent *component
 
-	// parent is the parent timer. It is used to progate the stop of the timer to the parent timer and get the remaining time in case the budget has to be inherited.
-	parent NodeTimer
+	// config is the configuration of the timer
+	config Config
+
+	// spent is the time spent on the timer, set after calling stop
+	spent time.Duration
 }
 
 var _ Timer = (*baseTimer)(nil)
--- a/timer/base_timer.go
+++ b/timer/base_timer.go
@@ -17,7 +17,7 @@ type baseTimer struct {
 	// config is the configuration of the timer
 	config config
 
-	timeCache
+	clock
 
 	// start is the time when the timer was started
 	start time.Time
@@ -25,6 +25,7 @@ type baseTimer struct {
 	// parent is the parent timer. It is used to progate the stop of the timer to the parent timer and get the remaining time in case the budget has to be inherited.
 	parent NodeTimer
 
+	// componentName is the name of the component of the timer. It is used to store the time spent in the component and to propagate the stop of the timer to the parent timer.
 	componentName string
 
 	// spent is the time spent on the timer, set after calling stop
@@ -45,8 +46,8 @@ func NewTimer(options ...Option) (Timer, error) {
 	}
 
 	return &baseTimer{
-		config:    config,
-		timeCache: newTimeCache(),
+		config: config,
+		clock:  newTimeCache(),
 	}, nil
 }
 
--- a/timer/base_timer.go
+++ b/timer/base_timer.go
@@ -13,6 +13,10 @@ import (
 // baseTimer is the type used for all the timers that won't have children
 // It's implementation is more lightweight than nodeTimer and can be used as a standalone timer using NewTimer
 type baseTimer struct {
+
+	// config is the configuration of the timer
+	config config
+
 	timeCache
 
 	// start is the time when the timer was started
@@ -21,12 +25,7 @@ type baseTimer struct {
 	// parent is the parent timer. It is used to progate the stop of the timer to the parent timer and get the remaining time in case the budget has to be inherited.
 	parent NodeTimer
 
-	// parentComponent is the component stored in the map of the parent timer. We keep a reference to it to update its spent time
-	// once the timer is stopped. It may be nil if the timer is not a child of a tree timer, like when calling NewTimer.
-	parentComponent *component
-
-	// config is the configuration of the timer
-	config config
+	componentName string
 
 	// spent is the time spent on the timer, set after calling stop
 	spent time.Duration
@@ -108,8 +107,7 @@ func (timer *baseTimer) Stop() time.Duration {
 
 	timer.spent = timer.Spent()
 	if timer.parent != nil {
-		timer.parentComponent.addSpent(timer.spent)
-		timer.parent.childStopped(timer.spent)
+		timer.parent.childStopped(timer.componentName, timer.spent)
 	}
 
 	return timer.spent
--- a/timer/base_timer.go
+++ b/timer/base_timer.go
@@ -1,0 +1,127 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2022 Datadog, Inc.
+
+package timer
+
+import (
+	"errors"
+	"time"
+)
+
+// baseTimer is the type used for all the timers that won't have children
+// It's implementation is more lightweight than nodeTimer and can be used as a standalone timer using NewTimer
+type baseTimer struct {
+	timeCache
+
+	// start is the time when the timer was started
+	start time.Time
+
+	// spent is the time spent on the timer, set after calling stop
+	spent time.Duration
+
+	// config is the configuration of the timer
+	config Config
+
+	// parentComponent is the component stored in the map of the parent timer. We keep a reference to it to update its spent time
+	// once the timer is stopped. It may be nil if the timer is not a child of a tree timer, like when calling NewTimer.
+	parentComponent *component
+
+	// parent is the parent timer. It is used to progate the stop of the timer to the parent timer and get the remaining time in case the budget has to be inherited.
+	parent NodeTimer
+}
+
+var _ Timer = (*baseTimer)(nil)
+
+// NewTimer creates a new Timer with the given options. You have to specify either the option WithBudget or WithUnlimitedBudget.
+func NewTimer(options ...Option) (Timer, error) {
+	config := newConfig(options...)
+	if config.Budget == InheritedBudget {
+		return nil, errors.New("root timer cannot inherit parent budget, please provide a budget using timer.WithBudget() or timer.WithUnlimitedBudget()")
+	}
+
+	if len(config.Components) > 0 {
+		return nil, errors.New("NewTimer: timer that have components must use NewTreeTimer()")
+	}
+
+	return &baseTimer{
+		config:    config,
+		timeCache: newTimeCache(),
+	}, nil
+}
+
+func (timer *baseTimer) Start() time.Time {
+	// already started before
+	if timer.start != (time.Time{}) {
+		return timer.start
+	}
+
+	if timer.config.Budget == InheritedBudget && timer.parent != nil {
+		timer.config.Budget = timer.parent.Remaining()
+	}
+
+	timer.start = timer.now()
+	return timer.start
+}
+
+func (timer *baseTimer) Spent() time.Duration {
+	// timer was never started
+	if timer.start == (time.Time{}) {
+		return 0
+	}
+
+	// timer was already stopped
+	if timer.spent != 0 {
+		return timer.spent
+	}
+
+	return time.Since(timer.start)
+}
+
+func (timer *baseTimer) Remaining() time.Duration {
+	if timer.config.Budget == UnlimitedBudget {
+		return UnlimitedBudget
+	}
+
+	remaining := timer.config.Budget - timer.Spent()
+	if remaining < 0 {
+		return 0
+	}
+
+	return remaining
+}
+
+func (timer *baseTimer) Expired() bool {
+	if timer.config.Budget == UnlimitedBudget {
+		return false
+	}
+
+	return timer.Spent() > timer.config.Budget
+}
+
+func (timer *baseTimer) Stop() time.Duration {
+	// If the current timer has already stopped, return the current spent time
+	if timer.spent != 0 {
+		return timer.spent
+	}
+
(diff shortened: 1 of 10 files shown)


timer/base_timer.go
```diff
@@ -0,0 +1,127 @@
...
+package timer
+
+import (
+	"errors"
+	"time"
+)
+
+// baseTimer is the type used for all the timers that won't have children
+// It's implementation is more lightweight than nodeTimer and can be used as a standalone timer using NewTimer
+type baseTimer struct {
```
RomainMuller, line 15: (Optional/naming) If they don't have children, they could be named `leafTimer`

timer/base_timer.go
```diff
@@ -0,0 +1,127 @@
...
+
+	// spent is the time spent on the timer, set after calling stop
+	spent time.Duration
+
+	// config is the configuration of the timer
+	config Config
+
+	// parentComponent is the component stored in the map of the parent timer. We keep a reference to it to update its spent time
+	// once the timer is stopped. It may be nil if the timer is not a child of a tree timer, like when calling NewTimer.
+	parentComponent *component
```
RomainMuller, line 29: (Nano-optimization) Suggest running `betteralign`... In particular, there is positive impact on time spent by GC on types if you keep all pointers at the start of the structure.

timer/base_timer.go
```diff
@@ -0,0 +1,127 @@
...
+// NewTimer creates a new Timer with the given options. You have to specify either the option WithBudget or WithUnlimitedBudget.
+func NewTimer(options ...Option) (Timer, error) {
+	config := newConfig(options...)
+	if config.Budget == InheritedBudget {
+		return nil, errors.New("root timer cannot inherit parent budget, please provide a budget using timer.WithBudget() or timer.WithUnlimitedBudget()")
+	}
+
+	if len(config.Components) > 0 {
+		return nil, errors.New("NewTimer: timer that have components must use NewTreeTimer()")
+	}
```
RomainMuller, line 46: Same here, I would prefer an API that does not allow mis-use. Run-time error detection should be a last resort. Ideally, if it compiles, it runs.

timer/base_timer.go
```diff
@@ -0,0 +1,127 @@
...
+
+	return &baseTimer{
+		config:    config,
+		timeCache: newTimeCache(),
+	}, nil
+}
+
+func (timer *baseTimer) Start() time.Time {
+	// already started before
+	if timer.start != (time.Time{}) {
```
RomainMuller, line 56: Curious why the parentheses there?

```suggestion
	if timer.start != time.Time{} {
```

timer/base_timer.go
```diff
@@ -0,0 +1,127 @@
...
+		return timer.start
+	}
+
+	if timer.config.Budget == InheritedBudget && timer.parent != nil {
+		timer.config.Budget = timer.parent.Remaining()
+	}
```
RomainMuller, line 62: Noice - capturing remaining time at start time... that's better than the previous iteration!

That said, I wonder if the `remaining` shouldn't be polled from the parent all the time instead of being copied down the tree at start time? That said, I'd worry this makes it difficult to reason over what concurrent usage means, and maybe that's a deal breaker.

timer/base_timer.go
```diff
@@ -0,0 +1,127 @@
...
+
+	remaining := timer.config.Budget - timer.Spent()
+	if remaining < 0 {
+		return 0
+	}
+
+	return remaining
+}
+
+func (timer *baseTimer) Expired() bool {
```
RomainMuller, line 95: Re-stating it there &mdash; but this is personal taste &mdash; I don't love `Expired` here because this actually refers to a budget having been consumed entirely, which I personally would call `Exhausted`.

timer/base_timer.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	return timer.spent
+}
+
+func (timer *baseTimer) Timed(timedFunc func(timer Timer)) (spent time.Duration) {
+	timer.Start()
+	defer func() {
+		timer.spent = timer.Stop()
+		spent = timer.spent
```
RomainMuller, line 122: Doesn't `timer.Stop()` set `timer.spent` and return its value?

```suggestion
		spent = timer.Stop()
```

timer/component.go
```diff
@@ -0,0 +1,34 @@
...
+	name string
+	// Updated each time a run of a sub-timer ends
+	spent atomic.Int64
+}
+
+// components store the data shared between child timers of the same component name
+type components struct {
+	lookup  map[string]*component
+	storage []component
+}
```
RomainMuller, line 21: (Micro-optimization, optional) I'm actually wondering if the GC would have less work to do if, instead of having `lookup map[string]*component` it was `lookup map[string]int` where the value is the index of the component in the `storage` map. That'd make fewer pointers for the GC to have to keep track of to determine reachability.

timer/component.go
```diff
@@ -0,0 +1,34 @@
...
+	storage := make([]component, len(names))
+	for i, name := range names {
+		storage[i].name = name
+		lookup[name] = &storage[i]
+	}
+	return components{
+		lookup:  lookup,
+		storage: storage,
+	}
+}
```
RomainMuller, line 34: Considering this usage, what's the added value of `storage`?

Feels like `components` could be `type components map[string]component`, and it'd have the same semantics, modulo ability to read the components in deterministic creation order; while incurring 1 allocation fewer?

timer/config.go
```diff
@@ -0,0 +1,64 @@
...
+import (
+	"time"
+)
+
+// UnlimitedBudget is a special value for the budget that means the timer has no budget
+var UnlimitedBudget = ^time.Duration(0) - 1
+
+// InheritedBudget is a special value for the budget that means the timer should inherit the budget from its parent
+// It is the default value if no options such as WithBudget, WithUnlimitedBudget or WithInheritedBudget are provided
+var InheritedBudget = time.Duration(0)
```
RomainMuller, line 17: Very optional... I wonder if its more clear that this is the 0-value of budget by making it:

```suggestion
var InheritedBudget time.Duration
```

timer/config.go
```diff
@@ -0,0 +1,64 @@
...
+// Copyright 2022 Datadog, Inc.
+
+package timer
+
+import (
+	"time"
+)
+
+// UnlimitedBudget is a special value for the budget that means the timer has no budget
+var UnlimitedBudget = ^time.Duration(0) - 1
```
RomainMuller, line 13: This is a pretty convoluted way to write `time.Duration(-2)`. And then why not `-1`?

I think it'd be best to use a more extreme value, as that'd stand out more when debugging. Since it's an `int64`, maybe the max `int64` as an approximation of "infinity":

```suggestion
var UnlimitedBudget = time.Duration(math.MaxInt64)
```

timer/config.go
```diff
@@ -0,0 +1,64 @@
...
+	Components []string
+}
+
+func newConfig(options ...Option) Config {
+	config := Config{}
+	for _, option := range options {
+		option(&config)
+	}
+	return config
+}
```
RomainMuller, line 33: TBH this type is easy enough to form by hand without using the constructor + option functors... and the zero-value would be what you expect it to be... and all its fields are exported, too...

You could imagine using the `Config` as a template for timers (that you could store in a package-level variable), and then use `Config.Start()` to create a new started timer with the designated budget & components (or `.New()` to create a new un-started timer with the designated budget & components).

timer/node_timer.go
```diff
@@ -0,0 +1,143 @@
...
+		},
+		components: newComponents(config.Components),
+	}, nil
+}
+
+func (timer *nodeTimer) NewNode(name string, options ...Option) (NodeTimer, error) {
+	config := newConfig(options...)
+	if len(config.Components) == 0 {
+		return nil, errors.New("NewNode: node timer must have at least one component, otherwise use NewLeaf()")
+	}
```
RomainMuller, line 53: This looks like a very artificial limitation, although I guess the return type difference justifies that...

timer/node_timer.go
```diff
@@ -0,0 +1,143 @@
...
+func (timer *nodeTimer) Stats() map[string]time.Duration {
+	stats := make(map[string]time.Duration, len(timer.components.lookup))
+	for name, component := range timer.components.lookup {
+		stats[name] = time.Duration(component.spent.Load())
+	}
+
+	return stats
+}
+
+func (timer *nodeTimer) SumSpent() time.Duration {
```
RomainMuller, line 116: This, and `SumRemaining` as well as `SumExpired` are confusing me. Why are those not what's returned by `Spent`/`Remaining`/`Expired`? To me, as a user, I'd only care about these `Sum` variants as the other forms would give me confusing/misleading information?

timer/time.go
```diff
@@ -0,0 +1,28 @@
...
+// Copyright 2022 Datadog, Inc.
+
+package timer
+
+import (
+	"time"
+)
+
+// timeCache is a simple cache for time.Now() to hopefully avoid some expensive calls to REALTIME part of time.Now()
+type timeCache struct {
```
RomainMuller, line 13: I would have called this `clock`, really.

timer/timer_test.go
```diff
@@ -0,0 +1,435 @@
...
+		require.GreaterOrEqual(t, newTimer.Spent(), time.Millisecond)
+		require.Equal(t, newTimer.Spent(), spent)
+	})
+}
+
+func TestSum(t *testing.T) {
+
+	if os.Getenv("GITHUB_RUN_ID") != "" {
+		t.Skip("Skipping test in CI environment")
+	}
```
RomainMuller, line 77: ... because ...?

I'll note that the `timeCache` / `clock` thing could be an interface that exposes `now() time.Time`, which you could then have a "real" implementation and a "fake" implementation that you advance "manually" for tests like these.

timer/base_timer.go
eliottness: At one point is was indeed called leafTimer but I think that it makes more sense, since this is also embedded into `nodeTimer` and used standalone-ly in `NewTimer`, to call it baseTimer. Since it's optional, ~closing~

timer/base_timer.go
eliottness: Benchmarks says we won ~5% perfs. Commiting this :+1:

timer/base_timer.go
```diff
@@ -0,0 +1,127 @@
...
+	parentComponent *component
+
+	// parent is the parent timer. It is used to progate the stop of the timer to the parent timer and get the remaining time in case the budget has to be inherited.
+	parent NodeTimer
+}
+
+var _ Timer = (*baseTimer)(nil)
+
+// NewTimer creates a new Timer with the given options. You have to specify either the option WithBudget or WithUnlimitedBudget.
+func NewTimer(options ...Option) (Timer, error) {
```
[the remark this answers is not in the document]
eliottness: The issue I have with this is that this would give a proactive use to the `UnlimitedBudget` special value and it is quite like an implementation detail (that I had to export because of the behaviour of `Remaining()` but still). I would prefer to have `NewTimerWithUnlimitedBudget` function but, for the same reason as I don't want a `NewContextWithBudget`, I don't feel that this is futur proof and the Option pattern has strictly no performance impact so :shrug:

timer/base_timer.go
eliottness: It does not compile without them. I think that go parser is greedily parsing the `{}` as the if it was the if block

timer/base_timer.go
```diff
@@ -0,0 +1,127 @@
...
+
+func (timer *baseTimer) Stop() time.Duration {
+	// If the current timer has already stopped, return the current spent time
+	if timer.spent != 0 {
+		return timer.spent
+	}
+
+	timer.spent = timer.Spent()
+	if timer.parent != nil {
+		timer.parentComponent.spent.Add(int64(timer.spent))
```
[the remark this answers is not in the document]
eliottness: Done :+1:

timer/base_timer.go
eliottness: Done :+1:

timer/component.go
```diff
@@ -0,0 +1,34 @@
...
+
+package timer
+
+import "sync/atomic"
+
+type component struct {
+	// name through which the component is identified
+	name string
+	// Updated each time a run of a sub-timer ends
+	spent atomic.Int64
```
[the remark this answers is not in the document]
eliottness: This value is quite frequently casted to `time.Duration` which is an alias type of `int64`

timer/component.go
eliottness: I tried what you are describing along a few other options (like only using a map without a slice) and this is the best one right now. Note that all the benchmarks spend more than 90% of their time in the `time` package anyway...

timer/component.go
eliottness: What you are describing does not work. doing a call to `lookup[name]` would copy the atomic around I would not be able to sum anything. If I do what you asked I have to have a pointer pointing to the atomic somewhere. Moreover benchmarks tell me that doing only a call to `make` is better performance-wise than doing 5 or 10 calls to `new`. 

But this made me think and I trying do simplify it a bit further.

timer/config.go
eliottness: I forgot to put it to const. And const have to be assigned :(

timer/config.go
eliottness: Ok so goland was broken I thought this value did not exist

timer/config.go
eliottness: Again I tried directly passing Config objects around but this did not end well and I had to modify 20 different tests to make things works again each time. Plus this is way easier for deprecation to works with the option system (that does not have any performance code as of right now)

timer/time.go
eliottness: I'll take it

timer/timer_test.go
eliottness: > ... because ...?
Because I still have to test the real implementation at some point even if with most tests I have ways for the tests to not be flaky.

Yes but I did not create a fake clock and fake timer for now because I don't want the PR to be bigger than this. And if I create a fake clock / timer I would also have to test it carefully before using it to test other stuff.

timer/base_timer.go
RomainMuller: ## LOL, WHAT??

timer/timer_test.go
RomainMuller: The spirit of that:

>... because ...?

Was to ask you to explain WHY you're skipping this in the CI environment in a comment or within the skip message, so it's easier for the next engineer around there to determine if this is still relevant or not ;)

eliottness: Reworked the test to make it work on the CI

PR #82, opened 2024-03-13
refactor: Stats & Metrics

eliottness:
- [x] Create the `Stats` structs to output structured metrics convertible into key-values ready for the backend
- [x] Store truncations in the code because there will be output merged between runs


encoder.go
```diff
@@ -56,6 +56,19 @@ const (
...
+func (reason TruncationReason) String() string {
+	switch reason {
+	case ObjectTooDeep:
+		return "depth"
+	case ContainerTooLarge:
+		return "container-size"
+	case StringTooLong:
+		return "string-size"
+	default:
+		return ""
```
RomainMuller, line 68: Not a fan of that kind of `Stringer` returning blank... Either this returns something "somewhat helpful, albeit not pretty", or it panics.

```suggestion
		return fmt.Sprintf("TruncationReason<0x%02X>", reason)
```

encoder.go
```diff
@@ -56,6 +56,19 @@ const (
...
 )
 
+func (reason TruncationReason) String() string {
+	switch reason {
+	case ObjectTooDeep:
+		return "depth"
+	case ContainerTooLarge:
+		return "container-size"
+	case StringTooLong:
+		return "string-size"
```
RomainMuller, line 66: (Optional) I wonder if I'd expect those to actually be "the constant name"...

context.go
```diff
@@ -301,58 +311,18 @@ func (context *Context) TotalTimeouts() uint64 {
...
 
-// merge merges the current metrics with new ones
-func (metrics *metricsStore) merge(other map[string]time.Duration) {
-	metrics.mutex.Lock()
-	defer metrics.mutex.Unlock()
-	if metrics.data == nil {
-		metrics.data = make(map[string]time.Duration, 5)
+	truncations := make(map[TruncationReason][]int, len(context.truncations))
+	for reason, counts := range context.truncations {
+		truncations[reason] = append([]int(nil), counts...)
```
RomainMuller, line 320: Smells like this could result in silly allocation patterns?

```suggestion
		truncations[reason] = make([]int, len(counts))
		copy(truncations[reason], counts)
```

RomainMuller (approved):
I like moving truncation information out of the results struct and into the context... this sounds a lot more efficient to use downstream, and I wonder why it didn't occur to me as an option...

encoder.go
eliottness: Panic is not really an option... going with your suggestion

eliottness: image

PR #83, opened 2024-03-13
fix: pointer receiver on a copied struct

eliottness:
Discrepancies between the function returning the Stats struct and the Metrics() function


Hellzy (approved):
Is this about perfs or usage?

eliottness:
Usage because the Stats() function returns a copy

PR #87, opened 2024-03-25
fix: report _dd.appsec.waf.duration even when the WAF does not match on anything

eliottness:
- [x] Fix a bug that would make that the value of the metric `_dd.appsec.waf.duration` was zero because the time was never added to the struct
- [x] Put back the metrics in float64 to be backwards compatible (even though it should already work as it is)

### DIFF
--- a/metrics.go
+++ b/metrics.go
@@ -37,7 +37,7 @@ const (
 func (stats Stats) Metrics() map[string]any {
 	tags := make(map[string]any, len(stats.Timers)+len(stats.Truncations)+1)
 	for k, v := range stats.Timers {
-		tags[k] = uint64(v.Microseconds())
+		tags[k] = float64(v.Nanoseconds()) / float64(time.Microsecond) // The metrics should be in microseconds
 	}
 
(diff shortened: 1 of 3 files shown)


metrics.go
```diff
@@ -37,7 +37,7 @@ const (
 func (stats Stats) Metrics() map[string]any {
 	tags := make(map[string]any, len(stats.Timers)+len(stats.Truncations)+1)
 	for k, v := range stats.Timers {
-		tags[k] = uint64(v.Microseconds())
+		tags[k] = float64(v.Nanoseconds()) / float64(time.Microsecond) // The metrics should be in microseconds
```
RomainMuller, line 40: I'll note that making the division in int space should be faster... at the expense of losing fractionals (do we care?)

RomainMuller (approved):

metrics.go
eliottness: Thats why it was an int before. Be at this point I think that backwards compatibility takes priority

PR #89, opened 2024-04-08
refactor: go 1.20 spring cleanup

eliottness:
- [x] Replace the multierror package by `errors.Join` since go 1.20 is our minimal supported version now
- [x] Remove usage of `go.uber.org/atomic` in favor of `sync/atomic`

### DIFF
--- a/waf.go
+++ b/waf.go
@@ -6,6 +6,7 @@
 package waf
 
 import (
+	"errors"
 	"fmt"
 	wafErrors "github.com/DataDog/go-libddwaf/v2/errors"
 	"sync"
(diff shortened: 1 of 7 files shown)


waf.go
```diff
@@ -165,7 +163,7 @@ func (r *Result) HasActions() bool {
 // Otherwise, it returns false along with an error detailing why.
 func SupportsTarget() (bool, error) {
 	wafSupportErrors := support.WafSupportErrors()
-	return len(wafSupportErrors) == 0, multierror.Append(nil, wafSupportErrors...).ErrorOrNil()
+	return wafSupportErrors == nil, errors.Join(wafSupportErrors...)
```
RomainMuller, line 166: I feel like these are making a somewhat dodgy assumption that `WafSupportErrors()` will return a nil slice if there is no error, when an empty slice would be semantically equivalent...

```suggestion
	err := errors.Join(support.WafSupportErrors())
	return err == nil, err
```

RomainMuller (approved):

waf.go
eliottness: errors.Join will return nil in both cases

PR #90, opened 2024-04-24
(chore): rework NewContext and NewContextWithBudget

Hellzy:
Part of APPSEC-52238

- Add an error return value
- Make NewContext and NewContextWithBudget handle methods


RomainMuller (approved):

eliottness:
Part of this task was to make `NewContext` a method of `Handle` it is still planned ?

Hellzy:
Is that what is meant by "rework NewContext" in the bucket list? I can add it there

PR #92, opened 2024-04-25
(chore): upgrade libddwaf to v1.17

Hellzy:
JIRA: APPSEC-52726

## Changes

- The WAF result actions are now a mapping from action type to action parameters, instead of a slice of action IDs
- Added `skip-smoke-tests` label for such API breaking changes where smoke-tests are expected to fail


context.go
```diff
@@ -276,9 +276,7 @@ func unwrapWafResult(ret bindings.WafReturnCode, result *bindings.WafResult) (re
 		return res, err
 	}
 	if size := result.Actions.NbEntries; size > 0 {
-		// using ruleIdArray cause it decodes string array (I think)
-		res.Actions, err = decodeStringArray(&result.Actions)
-		// TODO: use decode array, and eventually genericize the function
+		res.Actions, err = decodeMap(&result.Actions)
```
eliottness, line 279: So we output an "any" map as value? Can we do better ? like having a `map[string]map[string]string` maybe ?

Hellzy: I don't think specifying on string is a good idea since technically other types can be used for parameters (granted currently that's not the case). We could use `map[string]map[string]any` and make a round of cast after decoding like so:
```
	if size := result.Actions.NbEntries; size > 0 {
		m, err := decodeMap(&result.Actions)
		if err != nil {
			return res, err
		}
		res.Actions = make(map[string]map[string]any, len(m))
		for k, v := range m {
			val, ok := v.(map[string]any)
			if !ok {
				return res, errors.ErrInvalidObjectType
			}
			res.Actions[k] = val
		}
	}
```

I don't think it helps much though.

PR #98, opened 2024-05-14
feat: add actions diagnostics

Hellzy:


eliottness (approved):
can we have a test on this or is this too far-fetched?

PR #99, opened 2024-05-14
fix: stop passing empty strings as null ptrs

Hellzy:
Empty strings would be passed as null waf string objects, which the WAF doesn't handle in all cases.
While we wait for a WAF update to properly handle this, we set the waf object value for empty go strings to a null terminated empty C string representation.
This is band-aid solution that should get discarded once the WAF handles nullptr strings in all expected cases.

### DIFF
--- a/cgo_ref_pool.go
+++ b/cgo_ref_pool.go
@@ -27,6 +27,11 @@ type cgoRefPool struct {
 	arrayRefs  [][]bindings.WafObject
 }
 
+// This is used when passing empty Go strings to the WAF in order to avoid passing null string pointers,
+// which are not handled by the WAF in all cases.
+// FIXME: to be removed when the WAF handles null ptr strings in all expected places
+var emptyWAFStringValue = unsafe.NativeStringUnwrap("\x00").Data
+
 func (refPool *cgoRefPool) append(newRefs cgoRefPool) {
 	refPool.stringRefs = append(refPool.stringRefs, newRefs.stringRefs...)
 	refPool.arrayRefs = append(refPool.arrayRefs, newRefs.arrayRefs...)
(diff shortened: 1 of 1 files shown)


cgo_ref_pool.go
```diff
@@ -48,15 +48,17 @@ func (refPool *cgoRefPool) AllocCString(str string) uintptr {
...
-		return
+	if emptyStr {
+		str = "\x00"
 	}
 
 	refPool.stringRefs = append(refPool.stringRefs, str)
 	stringHeader := unsafe.NativeStringUnwrap(str)
+	if emptyStr {
+		stringHeader.Len = 0
+	}
```
eliottness, line 61: 🤔 Since strings in the WAF are not null-terminated, I am wondering how this is different than before. Sure now `obj.Value` is not null but noone should read any bytes of this pointer because it have a length of zero.

What error is this code fixing exactly?

Hellzy: What happens atm is that any empty string `""` gets converted to a null string pointer waf object, but the WAF in some cases only expects empty C strings OR no data at all, but not a null pointer. Such a case happens for the redirect action location parameter, for instance.
This code takes care of passing those empty Go string as empty C strings (so nil terminated).

Hellzy: Summarising our offline discussion:
- We don't want to have to pass null terminated strings in some cases since it breaks a big WAF precedent
- We expect the WAF to handle this case properly and will wait for an update to have a clean fix
- Meanwhile, we'll tweak the WAF objects in the case of empty strings so that they don't get tagged as null pointers on the WAF side.

eliottness (approved):

PR #103, opened 2024-07-03
compress `internal/lib/*.{so,dylib}` files when embedded and uncompress when loading to disk

paulcacheux:
Those embedded files represent a real weight that is impacting the serverless agent (where every MiB impacts the latency of the serverless). This PR compresses the files (so that the embedded version is stored compressed in the binary), and uncompresses them in `DumpEmbeddedWAF`


eliottness (approved):

RomainMuller (approved):

RomainMuller:
Noice! I must admit this has been floating in the back of my mind for ever, so thank you for actually going on an doing it :)

PR #104, opened 2024-07-04
refactor: set go 1.21 as minimal supported version

eliottness:
- [x] Upgrade go mod to 1.21
- [x] Remove go 1.20 from CI
- [x] use `errors.Join` when necessary
- [x] use builtin `max` when necessary

### DIFF
--- a/.github/workflows/_test_bare_metal.yml
+++ b/.github/workflows/_test_bare_metal.yml
@@ -8,7 +8,7 @@ jobs:
       fail-fast: false
       matrix:
         runs-on: [ macos-14, macos-13, macos-12, ubuntu-22.04, ubuntu-20.04, windows-latest, arm-4core-linux ]
-        go-version: [ '1.22', '1.21', '1.20' ]
+        go-version: [ '1.22', '1.21' ]
         include:
           # Test with DD_APPSEC_WAF_LOG_LEVEL (only latest go version)
           - go-version: '1.22'
--- a/internal/bindings/waf_dl.go
+++ b/internal/bindings/waf_dl.go
@@ -44,7 +44,6 @@ type wafSymbols struct {
 // The caller is responsible for calling wafDl.Close on the returned object once they
 // are done with it so that associated resources can be released.
 func NewWafDl() (dl *WafDl, err error) {
-	var handle uintptr
 	file, err := lib.DumpEmbeddedWAF()
 	if err != nil {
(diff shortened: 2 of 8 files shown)


.github/workflows/_test_bare_metal.yml
```diff
@@ -8,7 +8,7 @@ jobs:
       fail-fast: false
       matrix:
         runs-on: [ macos-14, macos-13, macos-12, ubuntu-22.04, ubuntu-20.04, windows-latest, arm-4core-linux ]
-        go-version: [ '1.22', '1.21', '1.20' ]
```
RomainMuller, line 11: No `1.23-rc` here?

internal/bindings/waf_dl.go
```diff
@@ -43,33 +44,25 @@ type wafSymbols struct {
...
-			} else {
-				// TODO: rely on errors.Join() once go1.20 is our min supported Go version
-				err = fmt.Errorf("%w; along with an error while removing %s: %v", err, file, rmErr)
-			}
+		if rmErr := os.Remove(file); rmErr != nil {
+			err = errors.Join(err, fmt.Errorf("error removing %s: %w", file, rmErr))
 		}
 	}()
 
-	var handle uintptr
```
RomainMuller, line 63: Why move this declaration higher up, when it's actually only useful starting there?

internal/lib/lib.go
```diff
@@ -28,20 +29,17 @@ func DumpEmbeddedWAF() (path string, err error) {
 	path = file.Name()
 
 	defer func() {
+		allErrors := []error{err}
```
RomainMuller, line 32: IIRC, you could just `err = errors.Join(err, newErr)` and spare the slice allocation?

RomainMuller (approved):
Nitpicks are take it or leave it :)

.github/workflows/_test_bare_metal.yml
eliottness: Nope, cf. [this](https://github.com/actions/go-versions/blob/main/versions-manifest.json)

internal/bindings/waf_dl.go
eliottness: Done

internal/lib/lib.go
eliottness: Done

PR #106, opened 2024-09-02
feat(lib): use memfd on linux instead of dumping libddwaf.so in /tmp

eliottness:
### What

This PR refactors the code in `internal/lib` to accommodate two implementations of the `DumpEmbeddedWAF()` function. One for linux using `memfd`, ultimately calling `dlopen("/proc/self/fd/{fd}")`, and one for darwin still creating a temporary file and calling `dlopen("/tmp/libddwaf-*.so")`.

Side quest: some function comments were not starting by the name of the function, fixed it.

### Motivation

Being able to use dd-trace-go on full read-only filesystems. On linux only for now but still


internal/lib/dump_waf_linux.go
```diff
@@ -0,0 +1,59 @@
...
+	"fmt"
+	"golang.org/x/sys/unix"
+	"io"
+	"os"
+
+	_ "embed"
+)
+
+//go:embed .version
+var EmbeddedWAFVersion string
```
Julio-Guerra, line 23: this could be deduplicated

internal/lib/dump_waf_linux.go
```diff
@@ -0,0 +1,59 @@
...
+
+func DumpEmbeddedWAF() (file *os.File, closer func() error, err error) {
+
```
Julio-Guerra, line 26: worth adding some comments about the background of this function

Julio-Guerra (changes requested):
Great idea 🎉

keisku (commented):
Can you update this documentation as well? Thank you!

https://github.com/DataDog/go-libddwaf/blob/48f7206417f9b0cbe3753d3792844c4bf3f07838/README.md?plain=1#L128-L137

internal/bindings/waf_dl.go
```diff
@@ -40,22 +40,22 @@ type wafSymbols struct {
...
 
-// newWafDl loads the libddwaf shared library and resolves all tge relevant symbols.
+// NewWafDl loads the libddwaf shared library and resolves all tge relevant symbols.
 // The caller is responsible for calling wafDl.Close on the returned object once they
 // are done with it so that associated resources can be released.
 func NewWafDl() (dl *WafDl, err error) {
-	file, err := lib.DumpEmbeddedWAF()
+	path, closer, err := lib.DumpEmbeddedWAF()
 	if err != nil {
 		return
```
keisku, line 49: Should we not ignore this `err`?

```suggestion
		return nil, fmt.Errorf("write an embedded WAF library: %w", err)
```

internal/bindings/waf_dl.go
```diff
@@ -40,22 +40,22 @@ type wafSymbols struct {
...
-			err = errors.Join(err, fmt.Errorf("error removing %s: %w", file, rmErr))
+		if rmErr := closer(); rmErr != nil {
+			err = errors.Join(err, fmt.Errorf("error removing %s: %w", path, rmErr))
 		}
 	}()
 
 	var handle uintptr
-	if handle, err = purego.Dlopen(file, purego.RTLD_GLOBAL|purego.RTLD_NOW); err != nil {
+	if handle, err = purego.Dlopen(path, purego.RTLD_GLOBAL|purego.RTLD_NOW); err != nil {
 		return
```
keisku, line 59:
```suggestion
		return nil, fmt.Errorf("dlopen: %w", err)
```

keisku (commented):
I don't have background context of errors that have been ignored.
Logging or returning an error should help support team in the future!

Julio-Guerra (approved):

internal/bindings/waf_dl.go
eliottness: Ah sh*t. Should have disabled auto-merge when I saw your comments. Don't worry I will do it in another PR

PR #109, opened 2024-09-11
feat: WAF Run Scope support for RASP Metrics

eliottness:
- [x] Add the `Scope` system to classfiy Waf Runs
- [x] Refactor Stats and Metrics without breaking the current API
- [x] Remove the `_dd.appsec` to all metrics for the futur when we support telemetry metrics

### DIFF
--- a/metrics.go
+++ b/metrics.go
@@ -48,7 +48,8 @@ const (
 	wafTruncationTag = "truncations"
 )
 
-// Metrics transform the stats returned by the WAF into a map of key value metrics for datadog backend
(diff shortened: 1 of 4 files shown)


metrics.go
```diff
@@ -53,20 +81,24 @@ type metricsStore struct {
 	mutex sync.RWMutex
 }
 
-func (metrics *metricsStore) add(key string, duration time.Duration) {
+func key(scope Scope, component string) string {
+	return fmt.Sprintf("%s.%s", scope, component)
```
Julio-Guerra, line 85: Replace by `strings.Builder`

Julio-Guerra (approved):

PR #120, opened 2025-04-24
fix: libddwaf does not accept nullptr-valued blank strings

RomainMuller:
Inadvertently replaced the pointer to a `\x00` byte that was used when encoding a blank string with a `0` (`null`) pointer; which `libddwaf` does not currently accept.

This commit restores the previous encoding behavior, and adds a test that was created based on the ruleset that allowed discovery of this issue.

---

Also fixes up the CGO-exported symbol name to be at V4 so it does not clash with the one exported by V3.


encoder.go
```diff
@@ -191,6 +194,19 @@ func (encoder *encoder) encode(value reflect.Value, obj *bindings.WAFObject, dep
 
 	//		Strings
 	case kind == reflect.String: // string type
+		if value.Type() == jsonNumberType {
```
eliottness, line 197: Can you add a case statement specifically for it

eliottness (approved):

PR #122, opened 2025-04-30
feat: remove timeout metrics and rework timer usages

eliottness:
This PR reworks go-libddwaf context and run API to better work with timers. It also removes timeout summing that is now done on dd-trace-go side.

- [x] Change `NewContext()` so it takes a list of timer options instead of simply the time budget for this context.
- [x] Add a `TimerKey` to the parameters passed to `Context.Run()` to select the component where this run duration will be summed in (ex: rasp or waf)
- [x] Create 3 public constants `EncodeTimeKey`, `DecodeTimeKey` and `DurationTimeKey` that all represent a part of a single WAF run that are supposed to be used as key to the map `Result.TimerStats` that was added as return value to `Context.Run()`.
- [x] Rework tests to they work with the changes
- [x] Adapt the README.md file to surface the main changes to the API
- [x] Remove metrics.go


RamyElkest (approved):
lgtm ⏲️

PR #123, opened 2025-05-07
feat: support multiple encoders

eliottness:
- [x] Create a new `Encoder` interface matching the current `encoder` struct already implementing it
- [x] Create the `NewEncoder` delegate type that can create any encoder.
- [x] Make the current `new*Encoder` function match the `NewEncoder` lambda type. Fuse those into a new public function `NewDefaultEncoder`
- [x] Add a `Encoder` option in the new WAF builder to pass it to the handle and then to the context for use
- [x] Adapt the code to use the new encoder lambda instead of the static new encoder function

### DIFF
--- a/builder.go
+++ b/builder.go
@@ -20,7 +20,8 @@ import (
 // or similar when sharing it across multiple goroutines. All methods of this
 // type are safe to call with a nil receiver.
 type Builder struct {
-	handle bindings.WAFBuilder
+	handle     bindings.WAFBuilder
+	newEncoder NewEncoder
 }
 
 // NewBuilder creates a new [Builder] instance. Its lifecycle is typically tied
@@ -41,7 +42,7 @@ func NewBuilder(keyObfuscatorRegex string, valueObfuscatorRegex string) (*Builde
 		return nil, errors.New("failed to initialize the WAF builder")
 	}
 
-	return &Builder{handle: hdl}, nil
+	return &Builder{handle: hdl, newEncoder: NewDefaultEncoder}, nil
 }
 
 // Close releases all resources associated with this builder.
@@ -72,7 +73,11 @@ func (b *Builder) AddOrUpdateConfig(path string, fragment any) (Diagnostics, err
 	var pinner runtime.Pinner
 	defer pinner.Unpin()
 
-	encoder := newMaxEncoder(&pinner)
+	encoder, err := b.newEncoder(newMaxEncoderConfig(&pinner))
+	if err != nil {
+		return Diagnostics{}, fmt.Errorf("could not create encoder: %w", err)
+	}
+
 	frag, err := encoder.Encode(fragment)
 	if err != nil {
 		return Diagnostics{}, fmt.Errorf("could not encode the config fragment into a WAF object; %w", err)
@@ -118,6 +123,17 @@ func (b *Builder) ConfigPaths(filter string) []string {
 	return wafLib.BuilderGetConfigPaths(b.handle, filter)
 }
 
+// Encoder takes [NewEncoder] function to be used in the handle created by the builder.
+// This function is not thread-safe and should be called before the builder is
+// built. The default encoder is [NewDefaultEncoder].
+func (b *Builder) Encoder(newEncoder NewEncoder) {
+	if b == nil || b.handle == 0 || newEncoder == nil {
+		return
+	}
+
+	b.newEncoder = newEncoder
+}
+
 // Build creates a new [Handle] instance that uses the current configuration.
 // Returns nil if an error occurs when building the handle. The caller is
 // responsible for calling [Handle.Close] when the handle is no longer needed.
@@ -132,5 +148,5 @@ func (b *Builder) Build() *Handle {
 		return nil
 	}
 
-	return wrapHandle(hdl)
+	return wrapHandle(hdl, b.newEncoder)
 }
--- a/encoder.go
+++ b/encoder.go
@@ -21,7 +21,43 @@ import (
 	"github.com/DataDog/go-libddwaf/v4/waferrors"
 )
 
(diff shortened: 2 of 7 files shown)


builder.go
```diff
@@ -72,7 +72,11 @@ func (b *Builder) AddOrUpdateConfig(path string, fragment any) (Diagnostics, err
 	var pinner runtime.Pinner
 	defer pinner.Unpin()
 
-	encoder := newMaxEncoder(&pinner)
+	encoder, err := newDefaultEncoder(newMaxEncoderConfig(&pinner))
```
RomainMuller, line 75: It's weird the `newDefaultEncoder` function accepts a configuration... It is no longer "default" I guess?

encoder.go
```diff
@@ -16,31 +16,35 @@ import (
...
+	// Pinner is used to pin the data referenced by the encoded wafObjects.
+	Pinner pin.Pinner
+	// Timer makes sure the encoder doesn't spend too much time doing its job.
+	Timer timer.Timer
+	// ContainerMaxSize is the maximum number of elements in a container (list, map, struct) that will be encoded.
+	ContainerMaxSize int
+	// StringMaxSize is the maximum length of a string that will be encoded.
+	StringMaxSize int
+	// ObjectMaxDepth is the maximum depth of the object that will be encoded.
+	ObjectMaxDepth int
```
RomainMuller, line 33: Since we're changing, stylistically, I think I kind of would prefer:

- `MaxContainerSize`
- `MaxStringSize`
- `MaxObjectDepth`

So basically the `Max` is always a prefix... it "feels" more coherent to me? Optional, though (the current names aren't too bad either)

encoder.go
```diff
@@ -16,31 +16,35 @@ import (
...
 // the encoded result. The Go references of the allocated wafObjects, along with every Go pointer they may
 // reference now or in the future, are stored and referenced in the `cgoRefs` field. The user MUST leverage
 // `keepAlive()` with it according to its ddwaf use-case.
 type encoder struct {
-	// pinner is used to pin the data referenced by the encoded wafObjects.
-	pinner pin.Pinner
-
-	// timer makes sure the encoder doesn't spend too much time doing its job.
-	timer timer.Timer
+	EncoderConfig
```
RomainMuller, line 44: Is it really desirable to allow this to be modified after creation?

encoder.go
```diff
@@ -78,28 +82,65 @@ const (
...
+type WAFObject = bindings.WAFObject
+
+// Encodable represent a type that can encode itself into a WAFObject.
+// The encodable is responsible for using the [pin.Pinner]
+// object passed in the [EncoderConfig] to pin the data referenced by the encoded [bindings.WAFObject].
+// The encoder must also use the [timer.Timer] passed in the [EncoderConfig] to
+// make sure it doesn't spend too much time doing its job.
+// The encoder must also respect the [EncoderConfig] limits and report truncations.
+type Encodable interface {
+	// Encode encodes itself as the WAFObject obj using the provided EncoderConfig and remaining depth allowed.
```
RomainMuller, line 96:
```suggestion
	// Encode encodes the receiver as a [WAFObject] using the provided [EncoderConfig] and remaining allowed depth.
```

encoder.go
```diff
@@ -78,28 +82,65 @@ const (
...
+	// It returns a map of truncation reasons and their respective actual sizes. If the error returned is not nil
+	// It is greatly advised to return errors from the waferrors package error when it matters.
```
RomainMuller, line 98: This feels like a mix of implementer notes & user documentation... I feel it'd be better to formulate everything as "user doc", while implying maintainer doc as a corrollary? Or vice-versa.

```suggestion
	// It returns a map of truncation reasons and their respective actual sizes. If the error returned is not nil,
	// it is greatly advised to return errors from the waferrors package error when it matters.
```

encoder.go
```diff
@@ -78,28 +82,65 @@ const (
...
+	// Outside of encoding the value, it is expected to check for truncations sizes as advised in the EncoderConfig
+	// and to regularly call the EncoderConfig.Timer.Exhausted() method to check if the encoding is still allowed
+	// and return waferrors.ErrTimeout if it is not.
+	// This method is NOT thread-safe.
```
RomainMuller, line 102: If this is implementer's doc, then "this method is not expected or required to be thread-safe". Also, Go typically states this as "safe/not safe to concurrently call from multiple goroutines"

encoder.go
```diff
@@ -78,28 +82,65 @@ const (
...
 
-func newMaxEncoder(pinner pin.Pinner) encoder {
-	timer, _ := timer.NewTimer(timer.WithUnlimitedBudget())
-	return encoder{
-		pinner:           pinner,
-		timer:            timer,
-		containerMaxSize: math.MaxInt,
-		stringMaxSize:    math.MaxInt,
-		objectMaxDepth:   math.MaxInt,
+func newMaxEncoderConfig(pinner pin.Pinner) EncoderConfig {
```
RomainMuller, line 138: This could be `newUnlimitedEncoderConfig` while the other one is just `newEncoderConfig` by contrast?

encoder.go
```diff
@@ -157,10 +188,18 @@ func isValueNil(value reflect.Value) bool {
 }
 
 func (encoder *encoder) encode(value reflect.Value, obj *bindings.WAFObject, depth int) error {
-	if encoder.timer.Exhausted() {
+	if encoder.Timer.Exhausted() {
 		return waferrors.ErrTimeout
 	}
 
+	if value.IsValid() && value.CanInterface() {
```
RomainMuller, line 195: `Value.CanInterface()` "returns whether `Interface` can be called without panicing", implying (and the implementations verify this) that `Value.IsValid()`, so there is no need to check both:


```suggestion
	if value.CanInterface() {
```

RomainMuller (approved):

PR #126, opened 2025-05-20
feat(encoder): identity-encode WAFObjects

RomainMuller:
This allows passing pre-converted objects directly to the WAF without addition effort incurred.

### DIFF
--- a/encoder.go
+++ b/encoder.go
@@ -82,9 +82,26 @@ const (
 	AppsecFieldTagValueIgnore = "ignore"
 )
 
-// WAFObject is the C struct that represents a WAF object. It is passed as-is to the C-world.
-// It is highly advised to use the methods on the object to manipulate it and not set the fields manually.
-type WAFObject = bindings.WAFObject
+type (
+	// WAFObject is the C struct that represents a WAF object. It is passed as-is to the C-world.
+	// It is highly advised to use the methods on the object to manipulate it and not set the fields manually.
+	WAFObject = bindings.WAFObject
+	// WAFObjectType describes the valid values of the [bindings.WAFObject.Type] field.
+	WAFObjectType = bindings.WAFObjectType
+)
+
+// Forwarding the constants from the bindings package so they can be re-used by external code that
+// produces pre-encoded [WAFObject] values.
+const (
+	WAFIntType    WAFObjectType = bindings.WAFIntType
+	WAFUintType   WAFObjectType = bindings.WAFUintType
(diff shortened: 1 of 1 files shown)


encoder.go
```diff
@@ -191,7 +208,17 @@ func (encoder *encoder) encode(value reflect.Value, obj *bindings.WAFObject, dep
...
-		if encodable, ok := value.Interface().(Encodable); ok {
+		iface := value.Interface()
+		if wo, ok := iface.(*bindings.WAFObject); ok {
+			*obj = *wo
+			return nil
+		}
+		if wo, ok := iface.(bindings.WAFObject); ok {
+			*obj = wo
+			return nil
+		}
```
eliottness, line 219: Wondering if this is even possible 🤔

How could we make sure that it works ?

Doing a value.Type() == wafObjectType could faster

encoder.go
```diff
@@ -82,9 +82,26 @@ const (
...
+const (
+	WAFIntType    WAFObjectType = bindings.WAFIntType
+	WAFUintType   WAFObjectType = bindings.WAFUintType
+	WAFStringType WAFObjectType = bindings.WAFStringType
+	WAFArrayType  WAFObjectType = bindings.WAFArrayType
+	WAFMapType    WAFObjectType = bindings.WAFMapType
+	WAFBoolType   WAFObjectType = bindings.WAFBoolType
+	WAFFloatType  WAFObjectType = bindings.WAFFloatType
+	WAFNilType    WAFObjectType = bindings.WAFNilType
+)
```
eliottness, line 104: I made sure there is no need to with methods on the WAFObject struct. Why do you need to access these ?

RomainMuller: There are useful when creating WAFObject values externally to this package, which I'd need to do in appsec-internal-go unless we move the default ruleset directly here... But we may still be interested in being able to do so in the future somewhere else (e.g, if we do JSON -> WAFObject directly and the decoder isn't in there).

encoder.go
RomainMuller: If I go `value.Type()` then I have to be ready to possibly recursively unwrap interfaces as well; which isn't done yet at this point... and is not necessary if the value is an `Encodable`...

Regarding "wondering if this is even possible" - I'm assuming you're referring to the interface having the non-pointer WAFObject type... And that is absolutely possible, in particular in the case of an `any` (but also in others) to wrap a value (by copy) in an eface.

PR #133, opened 2025-06-11
feat: expose encoder limits default values

e-n-0:
The PR promotes the encoder limits default values so they can be used to created default config for custom encoders.


eliottness:
These values are passed are part of a call to the `Encodable.Encode` in the `EncoderConfig` so there is no need for them for now

e-n-0:
Let's chat about that offline next time.

PR #136, opened 2025-06-18
feat(log): support for WAF logging on linux && !cgo

eliottness:
In #56 we argued that using `purego.NewCallback` was not safe enough to use on linux OS. A year and a half has passed and it can now be considered to be stable. Enable it

### DIFF
--- a/internal/log/log_purego.go
+++ b/internal/log/log_purego.go
@@ -16,17 +16,16 @@ import (
 )
 
 var (
-	once            sync.Once
+	once = sync.OnceValue(func() uintptr {
+		return purego.NewCallback(ddwafLogCallbackFn)
+	})
 	functionPointer uintptr
 )
 
 // CallbackFunctionPointer returns a pointer to the log callback function which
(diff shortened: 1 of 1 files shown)


internal/log/log_unsupported.go
```diff
@@ -3,7 +3,7 @@
 // This product includes software developed at Datadog (https://www.datadoghq.com/).
 // Copyright 2016-present Datadog, Inc.
 
-//go:build (!cgo && ((!darwin && !freebsd) || go1.26)) || datadog.no_waf
+//go:build (!linux && !darwin) || (!amd64 && !arm64) || go1.26 || datadog.no_waf || (!cgo && !appsec)
```
RomainMuller, line 6: TBH this could just be

```suggestion
//go:build !linux || !darwin || !amd64 || !arm64 || go1.26 || datadog.no_waf || !cgo || !appsec
```

RomainMuller (approved):

PR #137, opened 2025-06-23
chore: fix '-' serialization tag handling

RomainMuller:
The `-` exclusion is only when `"-"` is as-is, not when it has trailers (e.g, `"-,"`)


eliottness (approved):
Writing a test would be awesome

PR #145, opened 2025-08-07
DEBUG: log around purego calls (WARNING: MAKES THINGS SLOW)

RomainMuller:
This creates a new `$HOME/libddwaf-purego.<pid>.log` file at start up, and sync-writes to it about every call to `purego` being made (arguments and results). This intentionally calls `os.File.Sync` after each write and will hence make operations MUCH slower. Use only if instructed to.

### DIFF
--- a/internal/purego/purego.go
+++ b/internal/purego/purego.go
@@ -12,7 +12,6 @@ import (
 	"strings"
 	"sync"
 	"time"
-	"unsafe"
 
 	"github.com/ebitengine/purego"
 )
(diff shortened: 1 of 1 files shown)


internal/purego/purego.go
```diff
@@ -0,0 +1,84 @@
...
+	debug_log("Dlclose(%p)", handlePtr)
+	err := purego.Dlclose(handle)
+	debug_log("Dlclose(%p) = %v", handlePtr, err)
+	return err
+}
+
+func SyscallN(fn uintptr, args ...uintptr) (uintptr, uintptr, uintptr) {
+	var argsStr string
+	for _, arg := range args {
+		argsStr += fmt.Sprintf(", %p", unsafe.Pointer(arg))
```
eliottness, line 50: uintptr args may be simple int or flagset on some calls so are you sure no dereferencing is involved here ?

eliottness (changes requested):
Do you mind also logging when we create and delete the temp file libddwaf-xxx.dylib ? One of our hypothesis is that the delete operation on the file just after dlopen could have a negative impact

internal/purego/purego.go
RomainMuller: `%p` just prints `0x<address>`.

eliottness (approved):

PR #146, opened 2025-08-07
fix: dlopen'ing too many files causes kernel panic on M4 15.5

eliottness:
Use a common name to write the `libddwaf.dylib` and then dlopen it

### DIFF
--- a/internal/lib/dump_waf_darwin.go
+++ b/internal/lib/dump_waf_darwin.go
@@ -15,43 +15,98 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"path/filepath"
+	"syscall"
 )
+
+const libddwafDylibName = "libddwaf.dylib"
 
 // DumpEmbeddedWAF for darwin platform.
 // DumpEmbeddedWAF creates a temporary file with the embedded WAF library content and returns the path to the file,
 // a closer function and an error. This is the only way to make all implementations of DumpEmbeddedWAF consistent
 // across all platforms.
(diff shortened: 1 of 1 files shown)


internal/lib/dump_waf_darwin.go
```diff
@@ -15,43 +15,108 @@ import (
...
 // DumpEmbeddedWAF creates a temporary file with the embedded WAF library content and returns the path to the file,
 // a closer function and an error. This is the only way to make all implementations of DumpEmbeddedWAF consistent
 // across all platforms.
-func DumpEmbeddedWAF() (path string, closer func() error, err error) {
-	file, err := os.CreateTemp("", "libddwaf-*.dylib")
-	if err != nil {
-		return "", nil, fmt.Errorf("error creating temp file: %w", err)
-	}
+func DumpEmbeddedWAF() (_ string, closer func() error, err error) {
+	path := filepath.Join(os.TempDir(), libddwafDylibName)
```
RomainMuller, line 29: You'll want this to include some version string, too, or else all processes on the host MUST be using the same version.

RomainMuller (changes requested):

internal/lib/dump_waf_darwin.go
eliottness: done

RomainMuller (approved):
TBH this looks like a lot of complexity that might not be entirely warranted... Are we certain we want to merge this at this point?

